PR #13, opened 2019-05-29
Top the 'About' page

jpopelka:
I haven't found a better way how to make the About page the default one instead of the Posts.

The only way is to keep the date in the [about.md](https://github.com/packit-service/packit.dev/blob/master/content/about.md) newer than in the other content pages.


TomasTomecek:
it's possible we would had to tinker with the template.

Issue #27, opened 2019-07-12
documentation updates: intro to docs and P-a-a-S document move


TomasTomecek:
I think it would be more clear to our users if we:
* had the P-a-a-S document included within the docs/ tree (redirect the former URL if possible)
* and have a intro to the documention available at /docs/ - right now there is just a list of pages

PR #35, opened 2019-07-22
Packit-as-a-Service is on GitHub Marketplace. Fix docu

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

Since Packit-as-a-Service is available on GitHub Marketplace (https://github.com/marketplace/packit-as-a-service) we can use reference only to it.


phracek:
Make sense. We have the issue for it https://github.com/packit-service/packit.dev/issues/27

PR #61, opened 2019-10-09
packit principles: require green CI on master

TomasTomecek:


content/_index.md
```diff
@@ -76,6 +76,9 @@ We are also available on this e-mail: user-cont-team@redhat.com.
   every two weeks. To accomplish this, the release and validation processes are
   completely automated.
 
+* All the tests are passing in CI systems for master branches of every our
+  project. No excuses.
```
lachmanfrantisek: Isn't this a little bit negative?

TomasTomecek: please suggest better wording

TomasTomecek:
> What will be the best way how to achieve that?

Not merge code unless all is green :)

> I would like to clarify if we want our goal to be a green CI or unbreaked code...;-)

The goal is the first, and that should help us achieve the second things. Yes, tests can't catch all the bugs and it's up to us to do our best.

> What should we do if it breaks externally? (I am looking at you rebase-helper.)

Create an issue and fix it, as soon as possible. Should I add this to the text?

lachmanfrantisek:
> Create an issue and fix it, as soon as possible. Should I add this to the text?

I hope it's obvious...

content/_index.md
lachmanfrantisek: ` No excuses.` does not match the "serious" tone of the rest of the points...

TomasTomecek: YES!

TomasTomecek: Although, I can remove it :)

So, who wants me to remove it? :P

rpitonak: +1 for keeping it like this. I think it is worth highlighting, cause we (me including) have been merging without green CI a lot.

lachmanfrantisek: OK

We can make merging-via-zuul required to avoid merging PRs without green tests. (And ideally make te zuul check other checks.)

TomasTomecek: > OK
> 
> We can make merging-via-zuul required to avoid merging PRs without green tests. (And ideally, make the zuul check other checks.)

Actually zuul can do such a thing, I was told this at AnsibleFest: you can set up jobs which will require other GitHub checks to be passing (also got examples, hopefully)

jpopelka: Is "of every our project" grammatically correct?

TomasTomecek:
PTAL, I reworded based on the suggestion from @jpopelka

PR #68, opened 2019-11-18
Add a note about default files in synced_files

lbarcziova:
Related to packit-service/packit#603


jpopelka:
Will merge this after prod re-deploy

PR #83, opened 2020-01-09
properly render the content

TomasTomecek:
https://packit.dev/posts/January-20 did not work w/o this


TomasTomecek:
I needz this for da status reports.

PR #89, opened 2020-01-22
[Configuration] Update 'User configuration file'

### LINKED ISSUE #87: Outdated Configuration#User configuration file
[Minimal sample config](https://packit.dev/docs/configuration/#minimal-sample-config-1) says:

```yaml
debug: true
github_token: mnbvcxz123456
pagure_user_token: qwertyuiop098765
```

while I think it now should be
```yaml
debug: true

authentication:
  github.com:
    token: mnbvcxz123456
  pagure:
    token: qwertyuiop098765
    instance_url: https://src.fedoraproject.org
```

jpopelka:
Fixes #87


content/docs/configuration.md
```diff
@@ -256,31 +256,36 @@ The acceptable names are the same as for the package config:
  Key name                     | Type            | Description
 ------------------------------|-----------------|----------------------------------------------------------------------
  `debug`                      | bool            | enable debug logs
- `dry_run`                    | bool            | Do not perform any remote changes (pull requests or comments)
+ `dry_run`                    | bool            | do not perform any remote changes (pull requests or comments)
```
lachmanfrantisek: I am not sure it works... https://github.com/packit-service/packit/issues/232

content/docs/configuration.md
```diff
@@ -256,31 +256,36 @@ The acceptable names are the same as for the package config:
...
- `github_token`               | string          | Github API token: this is needed when packit interacts with Github API: https://github.com/settings/tokens (getting full read & write repo scope should be enough)
- `pagure_user_token`          | string          | Pagure token needed to access REST API, get it at: https://src.fedoraproject.org/settings#nav-api-tab
+ `authentication`             | dict            | tokens for services (GitHub, Pagure)
 
-Specifying tokens as environment variables `GITHUB_TOKEN` and
-`PAGURE_USER_TOKEN` is now deprecated and will be removed in future versions.
-Please store the tokens in the config file.
+The `authentication` is a dictionary where:
+* key is a hostname, url or name that can be mapped to a service-type, for example `github.com` or `pagure`
+* value is a dictionary with keys: `token`, `instance_url` (optional) and `type` (optional)
```
lachmanfrantisek: - `type` -- hard to explain that in few words
  - only if we cannot match the service-type from the service name.
  - if you use something like `github`/`pagure`/`gitlab` in the name, you don't need that

content/docs/configuration.md
```diff
@@ -256,31 +256,36 @@ The acceptable names are the same as for the package config:
...
+        token: qwertyuiop098765
+        instance_url: https://src.fedoraproject.org
+```
+
+The GitHub token is needed when packit interacts with Github API, get it at https://github.com/settings/tokens (getting full read & write repo scope should be enough).
+The Pagure token needed to access REST API, get it at https://src.fedoraproject.org/settings#nav-api-tab
+
+Specifying tokens as environment variables `GITHUB_TOKEN`/`PAGURE_USER_TOKEN` or
+as direct keys `github_token`/`pagure_user_token` is now deprecated and
+will be removed in future versions.
```
lachmanfrantisek: We already removed the support for the environment variables.

content/docs/configuration.md
jpopelka: OK, I think we can omit it for now since it's not in the example config below.

content/docs/configuration.md
jpopelka: removed

PR #92, opened 2020-01-26
polish configs, use aliases

TomasTomecek:
I dropped the listing b/c it is rendered incorrectly: https://packit.dev/docs/guide/


TomasTomecek:
> > I dropped the listing b/c it is rendered incorrectly
> 
> Seems rendered OK to me, but this also LGTM

it wasn't - the yaml indentation was wrong

PR #108, opened 2020-03-02
Be more explicit about who can trigger PaaS

csomh:
"project collaborators" was not an exact definition for who can trigger
Packit-as-a-Service. Make sure documentation is explicit about the
GitHub permissions required.

Partially fixes packit-service/packit-service#442.


content/packit-as-a-service/_index.md
```diff
@@ -59,7 +59,8 @@ Packit Service is able to schedule new builds based on a pull request comment:
 So whenever you run into a flake or feel like you want to retrigger, just type
 that comment into the PR and enjoy some fine, fresh builds.
 
-Only project collaborators can re-trigger.
+Only users with write or admin permissions to the repository can re-trigger
+COPR builds.
```
lachmanfrantisek: Don't we want to really use `collaborators`?

csomh: GitHub [terminology](https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/permission-levels-for-an-organization#outside-collaborators) calls "collaborators" users with *some kind* of access to an organization's repository, but not a member of the org. But this is not really relevant in this case. Both org members and collaborators might be missing write or admin permissions, in which case they won't be able to interact with Packit.

Put otherwise "collaborators" is a membership term, not a permission level one. Interacting with Packit requires the right permissions, there is no check for membership. The change in wording is to try to avoid this confusion :)

PR #138, opened 2020-06-29
propose-update: clean up, document tarball download

### LINKED ISSUE #137: document for propose-update that packit downloads the archive from Source(0)
> 16:51 <jkonecny> ttomecek, last thing, could you please mention somewhere in packit.dev that propose-update will download source from Source0
> 16:51 <jkonecny> I was missing this piece of information and I thought it will create a new archive like packit srpm

TomasTomecek:
Fixes #137


TomasTomecek:
asked twice for reviews, didn't get it, merging

jkonecny12:
Thanks a lot. It's much clearer now!

Just one small nitpick there is an issue with the `'` character in the ``--force`` option description on the https://packit.dev/docs/cli/propose-update/ (at least for Firefox 77.0.1). So maybe you want to fix this to make the web nicer. So or so, everything is written there and that is most important here ;).

TomasTomecek:
> Thanks a lot. It's much clearer now!
> 
> Just one small nitpick there is an issue with the `'` character in the `--force` option description on the https://packit.dev/docs/cli/propose-update/ (at least for Firefox 77.0.1). So maybe you want to fix this to make the web nicer. So or so, everything is written there and that is most important here ;).
> 
> P.S.: Sorry for my late reply. I was on vacation without internet connection last week.

Thanks for checking in.

Yeah, I saw the issue as well - tried to fix it but it couldn't; the only other option is to use a single color, which looked really blank, so I stuck to this one.

PR #155, opened 2020-08-19
source-git: document patch metadata

TomasTomecek:


content/docs/source-git/work-with-source-git.md
```diff
@@ -85,8 +85,52 @@ The important point with patch files is, that a commit which changes code (not
...
+
+#### Controlling the patch process
+
+With the process above you cannot name the patch file nor control where exactly
+should packit place the `Patch123: 123.patch` line in the specfile. There is a
+way though how you can do all of that yourself. You can set patch metadata in a
+commit message which packit will then read and take into account:
+* `patch_name` — name of the patch (e.g. `my-fancy.patch`)
+* `present_in_specfile:` — when set to `true`, it indicates the patch is set in
+  the spec file so packit should not touch the spec file
```
lachmanfrantisek, line 98: Don't we want put here all of them?

lachmanfrantisek (commented):
Thanks!

content/docs/source-git/work-with-source-git.md
TomasTomecek: why?

I'd say only these two make sense

lachmanfrantisek: To have it somewhere documented?

- It does not hurt us.
- People might want to know about others if they see it in the generated source-git. (e.g. `squash`)
- It can be useful for us as well. (I need to check code when I need that info.)

TomasTomecek: > To have it somewhere documented?
> 
>     * It does not hurt us.

It does when users get confused about the other 5 options which are not that clear. Even now I can see how many things in our project are super-confusing to users no matter how much docs we write.

>     * People might want to know about others if they see it in the generated source-git. (e.g. `squash`)

Yes, that's a good point, I'd still write it in the docs in a way that people are not meant to write those.

Anyway, what you are asking me here is dedicated docs for the patch metadata, I wanted this document to be focused on working with source-git - even when you read it the use case is pretty clear - if you wanna change patch name, this is how to do it. I'd like to see patch metadata documented somewhere else, fully, and probably link there from here.

>     * It can be useful for us as well. (I need to check code when I need that info.)

It can, we probably need documentation for us, since "us != our users".

TomasTomecek:
we have created a dedicated card for the complete documentation work, this should be good to merge, please approve

lachmanfrantisek (approved):

PR #157, opened 2020-08-25
faq: document how to reproduce sandbox env locally

TomasTomecek:
CC @jkonecny12


content/faq/_index.md
```diff
@@ -149,3 +149,53 @@ you need and us making them available for you.
...
+$ docker pull docker.io/usercont/sandcastle:prod
+```
+
+3. Launch the container and bind-mount the upstream project inside:
+
+```
+$ docker run -ti --rm -v $PWD:/anaconda docker.io/usercont/sandcastle:prod bash
+```
+
+4. Run commands of your choice:
```
lachmanfrantisek, line 175: It would be nice to have an image with both sandcastle dependencies and packit installed to be able to run `packit srpm` there as well.

content/faq/_index.md
```diff
@@ -149,3 +149,53 @@ you need and us making them available for you.
...
+checking for gcc... gcc
+checking whether the C compiler works... yes
+...
+```
+
+Our deployment is running in [OpenShift
+Online](https://www.openshift.com/products/online/) which is using docker as a
+container engine, that's why we are using docker here and not
+[podman](https://github.com/containers/podman) - if you wanna try podman,
+please make sure to run it as root.
```
jkonecny12, line 201: I don't really understand why it have to be run as root when the output of this is just to test srpm creation in that environment.

Also to be able to run ``podman`` with ``-v`` you have to use ``-v $PWD:/anaconda:Z`` the :Z is mandatory otherwise selinux will kill the mount.

jkonecny12 (commented):
Thanks a lot for this. It's great help!

One note below + I see that you have ``python3-pocketlint`` as dependency for Anaconda and Blivet-gui. It should not be needed anymore so you should be fine to remove this for SRPM archive creation.

content/faq/_index.md
jkonecny12: Seems that packit is already installed there.

lachmanfrantisek: Ah. Good to know..;)

content/faq/_index.md
TomasTomecek: thanks, I'm gonna add :Z, that's a good point

As for the root requirements: [rootless podman has its limitations](https://github.com/containers/podman/blob/master/rootless.md) and I'm personally tired off trying to figure those out or workaround them. Not even mentioning the performance hit b/c of the I/O being done in userspace via fuse - none of those happen when running as root.

csomh: @TomasTomecek could you share more details on the performance impact of using rootless podman you are seeing? I [was not able to reproduce anything significant](https://github.com/packit/dist-git-to-source-git/pull/31#discussion_r476283708).

podman is meant to be used in rootless mode, so I think a recommendation from our side to run it as root should be well explained, and the reasons given should be concrete and quantifiable.

content/faq/_index.md
```diff
@@ -149,3 +149,55 @@ you need and us making them available for you.
 
 In the mean time we are solving these requests one by one, so [please reach out
 to us](#how-can-i-contact-you).
+
+### A command failed in packit-service: how do I reproduce it locally?
+
+We don't have an end-to-end solution to this, yet.
+
+In the meantime, you can pull our production sandbox image
+(`docker.io/usercont/sandcastle:prod`) and run the command inside. As an
```
jpopelka, line 158: the exact image is specified below in `docker pull` & `docker run`, so I don't think it needs to be duplicated also here

```suggestion
and run the command inside. As an
```

jpopelka (approved):

content/faq/_index.md
TomasTomecek: Seems like I was wrong about the I/O: https://blog.tomecek.net/post/rootless-podman-io/

so I removed that part from the documentation

TomasTomecek:
recheck

PR #163, opened 2020-09-21
Update documentation for upstream_tag_template

sakalosj:
Gives this information any value to a user?

### DIFF
--- a/content/docs/configuration.md
+++ b/content/docs/configuration.md
@@ -87,6 +87,12 @@ Github repository.
 (*list of strings*) A command which prints current upstream version (hint:
 `git describe`) (defaults to `git describe --abbrev=0 --tags --match '*'`).
 
+This option affects the behavior of following CLI commands:
+* build
+* local-build
+* copr-build
+* srpm
+
 #### actions
 
 (*string*) Custom actions/hooks overwriting the default behavior of packit


content/docs/configuration.md
```diff
@@ -84,6 +84,11 @@ Github repository.
 (*list of strings*) A command which prints current upstream version (hint:
 `git describe`) (defaults to `git describe --tags --match '*.*'`).
 
+This option affects the behavior of following CLI commands:
+* build
+* local-build
+* copr-build
```
lachmanfrantisek, line 90: What about `srpm`?

sakalosj: I missed that, thanks 
fixed

content/docs/configuration.md
```diff
@@ -87,6 +87,12 @@ Github repository.
 (*list of strings*) A command which prints current upstream version (hint:
 `git describe`) (defaults to `git describe --abbrev=0 --tags --match '*'`).
 
+This option affects the behavior of following CLI commands:
+* build
```
lachmanfrantisek, line 91: Sorry for not mentioning that before, but `build` only checkout dist-git branch and trigger build in koji. (No `srpm` build is done.)

```suggestion
```

lachmanfrantisek (commented):
One more suggestion...;)

content/docs/configuration.md
sakalosj: correct me if I am wrong, but according to this

https://github.com/packit/packit/blob/e7c970d589bda74a3ad7d77acdd72c12717acb4f/packit/api.py#L419-L420

 it can affect also `build` command

TomasTomecek (approved):
LGTM

content/docs/configuration.md
lachmanfrantisek: Yes, you are right. Note that this is not a default behaviour but still influenced by that...

PR #169, opened 2020-10-01
Clarify how Packit Service requests permission to Copr projects

csomh:
Rephrase the relevant paragraphs in order to make it clear when Packit
Service asks for builder and admin permissions in custom Copr projects.


csomh:
I'll regenerate content once review is complete.

content/docs/configuration.md
```diff
@@ -318,12 +318,17 @@ Optional metadata:
...
   (`True` is `-1` and `False` is `60`).
 
-For custom `owner`, you will be asked to give us `builder` permission
-and in case of settings change (e.g. `targets`) also for `admin` permission.
-(You need to approve this request in the Copr project settings.)
-If you do not want to give us `admin` permission, you can do the change in Copr settings manually.
-Boolean values (`list_on_homepage` and `preserve_project`) are not updated
-when you use custom `owner`.
+When using a custom `owner`, Packit Service asks for `builder` permission the
+first time it tries to build in the project. In case the configuration of the
```
csomh, line 322: @lachmanfrantisek what is the workflow here? How will the jobs be restarted after the permissions were granted? I'm thinking maybe would be nice to clarify this part, too.

lachmanfrantisek: > what is the workflow here?

When we hit the error, we request for that permission and ask for a retrigger.

> How will the jobs be restarted after the permissions were granted?

User needs to restart that job.

- We don't know when/if we get that permission. (Maybe some periodic retrigger in the future?)
- For PRs, we clearly state that in [the build-related](https://github.com/packit/packit-service/blob/a16be74b3ed7f41abfdcee3da2901260ce0cdccc/packit_service/worker/build/copr_build.py#L390-L398) and [settings-related](https://github.com/packit/packit-service/blob/a16be74b3ed7f41abfdcee3da2901260ce0cdccc/packit_service/worker/build/copr_build.py#L344-L365) comment.
- There is no way ([for now](https://github.com/packit/packit-service/issues/780)) to retrigger the non-PR builds. (Only push will help.)
  - But they should receive a notification from Copr so some notification is there.

Your update is definitely much better but I am worried that it is not a cause of the confusion...;) (People will probably not read this before they hit the problem which is too later. But it's still good that we can link this part and not need to write that over and over.)

lachmanfrantisek (approved):

content/docs/configuration.md
csomh: Thank you for the explanation!

TomasTomecek (approved):

PR #174, opened 2020-10-19
week 42

TomasTomecek:


content/posts/October-2020.md
```diff
@@ -9,3 +9,9 @@ draft: false
 - [franta] made triggering of jobs more reliable and in correct situations ([packit-service#837](https://github.com/packit/packit-service/pull/837))
 
 [franta]: https://github.com/lachmanfrantisek
+
+## Week 42 (October 12th - October 16th)
+
+Nothing significant happened during this week. The only change was that Hunor
+[resolved](https://github.com/packit/packit-service/pull/850) a problem when sentry integration caused errors in stage
+environment and hence the problem didn't make it to production.
```
jpopelka, line 17: You'll hate me, but I'd skip the second sentence entirely. It was not a bug in our code, but in one of the dependencies we use and it was on stage (which implies that it "didn't make it to production").

csomh: Yeah, I agree, this is too internal. Maybe we should skip blogging on week 42? :)

TomasTomecek: I don't hate you, I :heart: you!

Okay, enough emotions - opinions. I understand your point, let me give mine: yes, I originally wanted to just write "Nothing significant" which says "yes, we are still alive thought no fixes nor RFEs" which in my head implies "we are lazy and didn't do anything for the whole week" and that's simply just wrong! it's just that none of our work is visible in this report for the last week.

With my current text I wanna message that "we are alive and we make sure that bugs don't hit prod" which to me is more interesting read than just "nothing". Also, it's just one semi-complex sentence, not a wall of text.

I will clarify in the text what sentry is and would prefer to leave it as it is.

Alternatively, I can post an update about upstream releases of packit (and ogr?).

csomh (approved):

PR #177, opened 2020-10-26
Document sync_changelog config option

lachmanfrantisek:


content/docs/configuration.md
```diff
@@ -49,6 +49,13 @@ archive.
 new pull request (defaults to `true`) or push directly to dist-git (if set to
 `false`).
 
+
+#### sync_changelog
+
+(*bool*) When doing a new update in Fedora dist-git, the spec-file changelog
+is synchronised when set to `true`. By default (`false`), everything but the changelog
+part is synchronized.
```
TomasTomecek, line 57:
```suggestion
part is synchronized. Use this only when your changelogs are in sync since this overwrites the changelog in the downstream.
```

TomasTomecek (changes requested):
we should warn people about this fact; also making sure this is explicitly stated since Fedora devs will hate this opt

lachmanfrantisek:
@TomasTomecek Can we merge? Do you want to add a link to the blog post?

TomasTomecek (approved):
thanks!

I already linked from the blog post

PR #178, opened 2020-10-30
Use Github Action to generate the pages

### LINKED ISSUE #78: Github Actions to generate/deploy site to GitHub Pages
Rebasing PRs are pain due to the generated HTML code.
If we could generate the result code after merging into master that would be much nicer.

There are already some existing Github Actions for this:
https://github.com/marketplace?type=actions&query=Hugo

jpopelka:
https://www.jameswright.xyz/post/deploy-hugo-academic-using-githubio/
https://github.com/marketplace/actions/hugo-deploy-github-pages

It's defined in
.github/workflows/deploy-pages.yml
and uses jpopelka-generated personal access token stored in
https://github.com/packit/packit.dev/settings/secrets

The generated content is being pushed to
https://github.com/packit/packit.dev-github-pages
from where the Github Pages are served.

Fixes #78


lachmanfrantisek (commented):
:tada: that's really nice! Thank you for looking into that!

A few questions:
- Is it serving from that `packit.dev-github-pages` already?
- What will happen when the generation fail? Or, is it possible?
  - We can add hugo generation to the pre-commit check if we want. (Just to check the ability to generate the content.) (ok, the versions of hugo can differ; it's not a good idea.)

lachmanfrantisek (approved):

jpopelka:
>     * Is it serving from that `packit.dev-github-pages` already?

Yes, it is. I pushed https://github.com/packit/packit.dev-github-pages/commit/b6460e0c2b564c062ffab675c32f9624430d9afc a while ago and you can see in bottom left of https://packit.dev that it is served from the new repo.



>     * What will happen when the generation fail? Or, is it possible?

I don't remember it failing so it's of course possible, but IMHO unlikely.

>     * We can add hugo generation to the pre-commit check if we want

#180

.github/workflows/deploy-pages.yml
```diff
@@ -0,0 +1,24 @@
+name: Build and Deploy
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
TomasTomecek, line 10: :<

.github/workflows/deploy-pages.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v1
+        with:
+          submodules: true
+
+      - name: Hugo Deploy GitHub Pages
+        uses: benmatselby/hugo-deploy-gh-pages@master
+        env:
+          HUGO_VERSION: 0.75.1
```
TomasTomecek, line 21: nice, we can actually select the precise version

TomasTomecek (approved):
nice!

.github/workflows/deploy-pages.yml
jpopelka: yeah, the [post I was following](https://www.jameswright.xyz/post/deploy-hugo-academic-using-githubio/) says: "The proceeding steps rely on access to apt-get to install the required software to build the static site files, so Ubuntu is chosen."

I don't know what steps it talks about so I left it unchanged.

PR #198, opened 2020-12-02
Github Action to check markdown links

### LINKED ISSUE #168: Test links
It would be nice to check the validity of the URLs in the documentation.

There is a tool called [liche](https://github.com/raviqqe/liche) that can do that for HTML and MD files. (Sadly, I am not aware of anything for Python.)

We can use that in pre-commit or as a [GitHub action](https://github.com/peter-evans/link-checker) like you can see e.g. [here](https://github.com/JackMcKew/awesome-python-bytes/blob/main/.github/workflows/schedule_link_checker.yml) (the action will create an issue with the report like [here](https://github.com/JackMcKew/awesome-python-bytes/issues/20))

jpopelka:
Fixes #168

(we'll see on Monday :-) )

### DIFF
--- a/.github/workflows/schedule_link_checker.yml
+++ b/.github/workflows/schedule_link_checker.yml
@@ -1,0 +1,21 @@
+name: Check markdown links
+on:
+  schedule:
+    # At 00:00 on Monday.
+    - cron: "0 0 * * 1"
+jobs:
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
Thanks!

PR #201, opened 2020-12-10
Don't run Github actions in forks of this repo

jpopelka:
Currently, they are run whenever I push to my master and fail because the secret is not configured in my fork.

I could disable Actions in my fork, but this commit should solve the problem for everyone.

https://github.community/t/stop-github-actions-running-on-a-fork/17965


TomasTomecek (approved):
LGTM

PR #215, opened 2021-01-07
update the official guide

TomasTomecek:
\+ document how people can use p-s with gitlab

### DIFF
--- a/content/docs/faq.md
+++ b/content/docs/faq.md
@@ -59,8 +59,14 @@ Packit connects the existing services ([Copr](https://copr.fedorainfracloud.org)
 
 ### Can we use Packit with Gitlab?
 
-Packit service is closely tied to GitHub right now because most of the projects are hosted on GitHub.
-[Let us know](https://github.com/packit/packit-service/issues/249) if you'd like to see Gitlab support.
+Yes!
+
+Since GitLab doesn't have an app functionality to enable integration, you need
+to manually configure the webhooks: hop into Settings → Webhooks → Add Webhook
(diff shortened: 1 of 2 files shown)


jpopelka (approved):

content/docs/faq.md
```diff
@@ -57,8 +57,14 @@ Packit connects the existing services ([Copr](https://copr.fedorainfracloud.org)
...
-Packit service is closely tied to GitHub right now because most of the projects are hosted on GitHub.
-[Let us know](https://github.com/packit/packit-service/issues/249) if you'd like to see Gitlab support.
+Yes!
+
+Since GitLab doesn't have an app functionality to enable integration, you need
+to manually configure the webhooks: hop into Settings → Webhooks → Add Webhook
+and enter "https://prod.packit.dev/api/webhooks/gitlab"
+
+Please bear in mind that not many people are using packit service via GitLab,
+so if anything doesn't work as expected, please reach out to us.
```
TomasTomecek, line 67: I'm 100% sure this is not enough. How does approval work here?

@lachmanfrantisek am I missing something?

lachmanfrantisek: There are two options:
* You have a secret token we have in our config file. (Probably not relevant to the user.)
* You will receive a secret token on the first event as a confidential issue. They just need to copy that value and place it as a secret token send with the webhook. (In the webhook settings.)

TomasTomecek: :+1: - I am missing something :D

TomasTomecek:
merging since this is better than saying up front we don't support gitlab - we do!

PR #225, opened 2021-01-13
document "Generated code in upstream archives"

TomasTomecek:
is this understandable?

### DIFF
--- a/content/docs/archive-not-matching-git.md
+++ b/content/docs/archive-not-matching-git.md
@@ -1,0 +1,80 @@
+---
+title: "Generated code in upstream archives"
+draft: false
+disableToc: false
+weight: 10
+---
+
+# When the release archive does not match the upstream git tag checkout
+
+This document covers a scenario when an upstream project has a script to create
+release archives and some of the code in the archive is generated. This means
+that when a downstream distribution has patches of the generated code those
+patches can't be applied in the upstream repo.
+
+
+## The solution
+
+The provided solution will be performed in a [source-git]({{< ref "source-git" >}})
+repository.
+
+When all your downstream patches apply cleanly on top of the upstream git repo
+for the particular git tag, then all is good and you don't need anything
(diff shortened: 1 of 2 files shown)


content/docs/archive-not-matching-git.md
```diff
@@ -0,0 +1,55 @@
...
+draft: false
+disableToc: false
+weight: 10
+---
+
+# When the release archive does not match the upstream git tag checkout
+
+This document covers a scenario an upstream project has a script to create
+release archives and some of the code in the archive is generated. This means
+that a downstream distribution mat then have downstream patches of the
```
voxik, line 12: s/mat/must/

content/docs/archive-not-matching-git.md
```diff
@@ -0,0 +1,55 @@
...
+## The solution
+
+The ideal solution is to go for [source-git]({{< ref "source-git" >}}).
+
+When all your downstream patches apply cleanly on top of the upstream git repo
+for the particular git tag, then all is good and you don't need anything
+special. You can follow [the guide for the init command]({{< ref "init" >}}) to
+create such a source-git repo.
+
+The real problem rises when you need to patch code which is **not** present in
```
voxik, line 25: I think the real problem (speaking in the context of Ruby) is the opposit case, i.e. there is upstream change, which I'd like to apply downstream, but it would need to went through the preproces script to be applied properly to the tarball. The issue is that I don't have the tools available, e.g. I'd need to have Ruby to be able to successfully execute the preprocess script.

TomasTomecek:
@voxik thanks for the review! I briefly talked to @kdudka yesterday who described the same issue to me with project using autotools. Sadly I don't think there is anything we can do to help - if you don't have the right tools in the downstream to regenerate the code, we can't help.

I added another section to the doc to describe this in detail.

lbarcziova (approved):
well written!

PR #226, opened 2021-01-14
faq: reproduce locally: update

TomasTomecek:
update the faq with more relevant info

document how to invoke packit in a worker container

Should slightly help with:
https://github.com/packit/packit-service/issues/917


jpopelka (approved):

content/docs/faq.md
```diff
@@ -210,3 +217,59 @@ Our deployment is running in [OpenShift
 Online](https://www.openshift.com/products/online/) which is using docker as a
 container engine, that's why we are using docker here and not
 [podman](https://github.com/containers/podman).
+
+If this is not enough, there is more! You can pull our production worker image
+and run packit commands in there to reproduce issues with parsing packit.yaml
+or interacting with the git repo.
+```
+$ docker pull docker.io/usercont/packit-worker:prod
+```
```
martinpitt, line 226: I'd drop that part, as podman/docker run pull the image automatically if necessary.

content/docs/faq.md
```diff
@@ -210,3 +217,59 @@ Our deployment is running in [OpenShift
...
+There are multiple operations which happen in the worker:
+* cloning repositories
+* parsing packit.yaml
+* interacting with the git repo
+* interacting with other services
+
+This how we can create a SRPM in a very similar way packit service does it (ogr
+is one of the libraries powering packit):
+```
+$ docker run -it --rm -v $PWD:/ogr -w ogr docker.io/usercont/packit-worker:prod bash
```
martinpitt, line 237: Suggestion: The "ogr" bit is not magic, right? So maybe more generally `-v $PWD:/src -w /src` to avoid confusion?

martinpitt (commented):
Thanks!

I tried this with cockpit-podman, and it immediately fails as the packit-worker container does not have npm:
```
❱❱❱ podman run -it --rm -v $PWD:/source -w /source docker.io/usercont/packit-worker:prod bash
bash-5.0# packit srpm
Failed to set locale, defaulting to C.UTF-8
Using user-defined script for ActionName.post_upstream_clone: [['make', 'cockpit-podman.spec']]
sed -e 's/%{VERSION}/27.1.8.g0a81eec/g' cockpit-podman.spec.in > cockpit-podman.spec
Using user-defined script for ActionName.create_archive: [['sh', '-ec', '(make dist-gzip 2>&1 </dev/null) | cat'], ['make', 'dist-gzip-name']]
# if it exists already, npm install won't update it; force that so that we always get up-to-date packages
rm -f package-lock.json
# unset NODE_ENV, skips devDependencies otherwise
env -u NODE_ENV npm install
env: ‘npm’: No such file or directory
make: *** [Makefile:199: package-lock.json] Error 127
/source/cockpit-podman-27.1.8.g0a81eec.tar.gz
ERROR    Preparing of the upstream to the SRPM build failed: The create-archive action did not output a path to the generated archive. Please make sure that you have valid path in the single line of the output. 
```

This trivial reason is obviously not what breaks [in production](https://prod.packit.dev/srpm-build/11988/logs), as there `npm` *is* available. I suppose that part really does run in a sandcastle container instead of a packit-worker container. So apparently the YAML parsing and the actual build happen in two different places. Maybe it's possible to run these steps separately locally as well?

Anyway, I have two minor suggestions, but this is a great help already!

TomasTomecek:
Thanks Martin! I'm gonna play with this a bit more today and will try to reproduce the cockpit-podman issue.

It's true that in production, those operations run in 2 different environments - back then I tried to implement "local containers as sandbox" for sandcastle but never finished it :/

TomasTomecek:
I dropped the worker part and replaced it with an actual python code how to run commands in openshift the same way packit-service does it.

martinpitt:
Perfect, thanks @TomasTomecek ! Indeed for me the `--memory` bit was the crucial thing. I can perfectly reproduce the failure with 512MB. I still get OOM with 768MB as in https://github.com/packit/sandcastle/pull/92, but it works with 1024MB. But as you added cockpit-podman to sandcastle's integration tests, I suppose that's fine.

However, I just ran https://github.com/cockpit-project/cockpit-podman/pull/649 again, and it [still failed](https://prod.packit.dev/srpm-build/12229/logs) pretty much the same way. https://github.com/packit/sandcastle/pull/92 is rolled out already, though. But I'll follow up in another issue/PR about that, not this one.

Thanks a lot for the Python reproducer, this is really useful!

martinpitt (approved):

PR #227, opened 2021-01-15
source-git: add an image with a real example

TomasTomecek:


csomh (approved):
:+1: though "Add a diagram showing an example of a source-git history" would probably express the goal of this commit better. I had to look at the diff to figure out that this is not about some *container image*.

PR #231, opened 2021-01-27
source-git: provide E2E guide

TomasTomecek:
TODO:
* [x] write the final chapter
* [x] add links to subsections from the root sg document
* [x] read through all again and polish


content/docs/source-git/_index.md
```diff
@@ -39,12 +40,13 @@ such a thing:
...
 ### Spec file
 
-I'm going to answer the elephant in the room right away: "Do we really need to
-have a spec file in our upstream repo or a source-git repo?!" — No, you don't,
-though storing the spec file directly in the repository is the most convenient.
-Alternatively, packit needs to have a way to obtain the spec file, so, you can:
+We are going to answer the elephant in the room right away: "Do we really need to
+[have a spec file in our upstream repo or a source-git repo?!]({{< ref "faq#how-can-i-download-rpm-spec-file-if-it-is-not-part-of-upstream-repository" >}})"
+No, you don't, though storing the spec file directly in the repository
+will save you a lot of trouble. Alternatively, packit needs to have a way to
```
csomh, line 46: What "trouble"? :)

Storing the spec file in the same tree as the source code creates a tight-coupling between the two. This might be something that is wanted or something that is not.

If the spec file is stored next to the source code and it's used for building the package, each build can be traced back to a given commit and so a given (source code, spec file) pair. It is obvious how to find the version of the spec file which was used.

If the spec file is stored some other place, the spec file can be changed without this change being reflected in the history of the source code. So tracing back a build to a given (source code, spec file) pair becomes problematic or would require additional tooling. Maybe this is the trouble you though of? :slightly_smiling_face:

csomh, line 46: Shouldn't we capitalise `Packit`? Written like this it can be confusing for readers who are already not familiar with it.

content/docs/source-git/_index.md
```diff
@@ -8,17 +8,18 @@ bookCollapseSection: true
...
 [dist-git](https://github.com/release-engineering/dist-git). The need was
 especially amplified after we realized that many projects already do this. The
 problem was that everyone created a workflow and tooling for themselves which
-made it really hard to share knowledge and onboard new projects. Our intention
+made it really hard to share knowledge and on-board new projects. Our intention
 is to create an implementation which is highly configurable, easy to start with
 and provides both command-line tooling and a service which is tightly
 integrated with GitHub and GitLab.
 
+> Before you continue reading, we expect you are familiar with [Fedora
```
csomh, line 20:
```suggestion
> Before you continue reading, we assume you are familiar with the [Fedora
```

content/docs/source-git/_index.md
```diff
@@ -58,6 +60,16 @@ any way you want:
 * Script it yourself
 
 
+### How packit treats `git commit`s
```
csomh, line 63: Hm, it's not clear why this topic is important. Could you explain more?

```suggestion
### How does Packit handle git commits?
```

content/docs/source-git/_index.md
```diff
@@ -58,6 +60,16 @@ any way you want:
...
+
+Packit distinguishes between 2 types of commits:
+1. Those that change code (usually pulled from upstream)
+2. Changes to packaging (spec file, other sources, changes to .packit.yaml)
+
+Packit treats code changes as downstream patches so it creates those files
```
csomh, line 69: Not sure I understand....

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+weight: 3
+bookCollapseSection: true
+---
+
+# Working with source-git
+
+This document is focused on how packit treats a source-git repo.
+
+If you are looking for the design concept of source-git, please head on here:
+[docs/source-git-design]({{< ref "/docs/source-git/design.md" >}})
```
csomh, line 14:
```suggestion
If you are looking for the design concept of source-git, please [read it here]({{< ref "/docs/source-git/design.md" >}}).
```

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+  tests on those builds, and getting feedback in pull (merge) requests.
+
+Let's describe all of these in detail.
+
+
+## Upstream git history or just an unpacked tarball?
+
+Your source-git repo can contain upstream git history if you want. Or it
+doesn't. The alternative is to unpack a tarball corresponding to an upstream
+release. This is really up to you and packit is able to work with both.
```
csomh, line 35: Do we want to promote this alternative?
How the history is going to look like. Is a source-git repo like this still useful for development? Does it considerably easier to maintain the package?

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+the git ref which is supposed to match the archive specified in the spec file.
+All commits on top of this ref are treated either as:
+
+* downstream patches
+* downstream packaging
+
+For downstream patches, packit creates a patch file for each commit on top of
+`upstream_ref`. Downstream packaging changes are ignored in patch files since
+they can be directly committed to dist-git. You can control the ignore
+mechanism with config option `patch_generation_ignore_paths`.
```
csomh, line 47: Link the docs of this config option.

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+* And you are able to utilize all features of packit-service: RPM builds of PRs,
+  tests on those builds, and getting feedback in pull (merge) requests.
+
+Let's describe all of these in detail.
+
+
+## Upstream git history or just an unpacked tarball?
+
+Your source-git repo can contain upstream git history if you want. Or it
+doesn't. The alternative is to unpack a tarball corresponding to an upstream
```
csomh, line 34: Wording feels sloppy, could you make it a little bit more concrete and structured?

Would it make sense to reference docs for creating such source-git repositories?

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+* 3ee90ce - the bottom commit, equals to an unpacked upstream tarball
+
+packit.yaml has only two lines:
+
+    $ cat .packit.yaml
+    specfile_path: SPECS/pacemaker.spec
+    upstream_ref: source-git-start
+
+As you can see, we are telling packit, that commit tagged as `source-git-start`
+indicates the barrier between upstream and downstream, in our case, it's just a
```
csomh, line 75: Not sure this belongs here, but couldn't this start be indicated in some other way? Cannot this be the name of the upstream branch, for example? Where I would like to get with this: could we remove the requirement of creating a tag to mark the border between upstream and downstream and use, for example, only the name of the upstream branch?

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+single commit with the upstream archive content, but it could have been the
+whole upstream git history.
+
+
+### Rebase or merge?
+
+A common workflow in the world of open source development is to accept
```
csomh, line 82: This sounds very fancy but it's very low on information :)

Could it be simplified?

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+pull requests and have merge commits in the git history. If the PRs are not
+rebased against the main branch before the merge, the history may go wild.
```
csomh, line 84: Could "go wild" replaced by a more technical/concrete expression?

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+Packit is able to work with such a state by creating an ephemeral branch with
+linear history and generate the patch files out of it. [More
+info](https://github.com/packit/packit/pull/766).
```
csomh, line 87: Shouldn't we write this up and create and article for it, instead of referencing a PR? Seeing a PR might scare readers away.

content/docs/source-git/work-with-source-git/_index.md
```diff
@@ -0,0 +1,112 @@
...
+best:
+
+1. **Rebase against the main upstream branch** (a.k.a. `git pull --rebase
+   upstream master`). This solution implies that you are able to perform the
+   force-push operation (e.g. Fedora and CentOS dist-git instances don't allow
+   force-pushes). Packit will not help you with the rebase process in any way
+   and it's up to you to complete the rebase.
+2. **Create a new branch for every upstream release**. No rebase needed — may
+   require good branch hygiene since the number of branches can skyrocket. No
+   dedicated help from packit again, you are running all the `git` commands.
```
csomh, line 100: Wondering if there would be any "hand-holding" that Packit could do here? :thinking:

content/docs/source-git/work-with-source-git/build-locally.md
```diff
@@ -0,0 +1,52 @@
...
+SRPM: /home/tt/g/systemd/systemd-stable/systemd-247.1-2.g68d22b32.fc33.src.rpm
+```
+
+Our changes worked! We can try building them, either in our local environment
+using packit - this implies that all build dependencies are installed:
+```
+$ packit local-build
+```
+
+...or in [mock](https://github.com/rpm-software-management/mock):
```
csomh, line 29: :thinking:

Would it make sense to wrap this into `packit local-build`?

content/docs/source-git/work-with-source-git/build-locally.md
```diff
@@ -0,0 +1,52 @@
...
+
+If the build is passing locally, it may be time to [propose your changes to
+dist-git]({{< ref "propose-to-dist-git.md" >}}).
+
+
+## How packit generates an SRPM from a source-git repo?
+
+These are the steps:
+
+1. Download archive specified in specfile's Source directive.
```
csomh, line 46: This always confuses me... why isn't the archive created from the Git repo the packit command is called in?

content/docs/source-git/work-with-source-git/build-locally.md
```diff
@@ -0,0 +1,52 @@
...
+2. Create patch files from commits on top off `upstream_ref` where necessary.
+3. Bump release in the spec file.
+4. Generate new changelog entry in the spec file.
```
csomh, line 49: Maybe link docs on how this can be configured.

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
```diff
@@ -0,0 +1,95 @@
...
+date: 2021-01-26T23:59:59+01:00
+draft: false
+weight: 2
+---
+
+# Pull fixes from the upstream
+
+One of the common tasks of being a downstream maintainer is to pull fixes from
+the upstream codebase. This starts by creating a new bugzilla with an issue in
+the distribution which requires a code-change in the affected package. The
```
csomh, line 12: Is this also required in Fedora? Maybe we should explicitly clarify who the intended audience is for these docs?

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
```diff
@@ -0,0 +1,95 @@
...
+maintainer needs to pull it from upstream when:
+1. The fix is not released yet by the upstream - this implies pulling the fix
+   from the main development upstream branch.
+2. If the fix is already released, one can either update to that upstream
+   release or only pull commits with the fix. Please consult [Fedora Packaging
+   Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/) if
```
csomh, line 18: Would it be possible to narrow the scope of this link? The Guidelines are pretty long, which section/s is/are relevant here?

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
```diff
@@ -0,0 +1,95 @@
...
+using https://gitlab.com/packit-service/src/systemd.
+```
+$ git remote -v
+origin      git@gitlab.com:TomasTomecek/systemd-stable.git (push)
+source-git  git@gitlab.com:packit-service/src/systemd.git (push)
+upstream    git@github.com:systemd/systemd-stable.git (push)
+```
+
+We have 3 remotes in the repo:
+1. `origin` — a fork of the source-git repo
```
csomh, line 39: Fork-remotes set up by `fedpkg fork` are called by the FAS username, not `origin`. Should we also follow this pattern in source-git?

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
```diff
@@ -0,0 +1,95 @@
...
+2. It also adds them into the spec file as new `PatchXYZ` entries.
+
+If that doesn't work for you, you can tailor the patch process. The main
+downside is that with the default process you cannot name the patch file nor
+control where exactly should packit place the `Patch123: 123.patch` line in the
+spec file. There is a way though how you can do all of that yourself. You can
+set patch metadata in a commit message which packit will then read and take
+into account:
+* `patch_name` — name of the patch (e.g. `patch_name: my-fancy.patch`)
+* `present_in_specfile:` — when set to `true`, it indicates the patch is set in
```
csomh, line 61: Is this required? What is the default?

csomh:
```suggestion
* `present_in_specfile` — when set to `true`, it indicates the patch is set in
```

csomh (commented):
Nice! :smiley_cat:

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
TomasTomecek: the default is "false" which means that packit will try to add the patch in the specfile. That may result into a borked spec file with one Patch being defined twice (or applied twice) - we could make packit smarter about this

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
TomasTomecek: sounds good! this was my random local setup when creating some source-git repo, will do!

also, I never used `fedpkg fork`

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
TomasTomecek: will do - the relevant section should be "don't change API/ABI in a stable Fedora release)

TomasTomecek: okay, I even linked to the wrong doc, thanks for the pointer!

content/docs/source-git/work-with-source-git/pull-upstream-fixes.md
TomasTomecek: Bugzilla is not required in Fedora to push an update. I dropped that info and simplified this paragraph.

I'm sorry but I don't understand where and how to write about the intended audience.

content/docs/source-git/work-with-source-git/build-locally.md
TomasTomecek: What exactly is meant to be configured?

content/docs/source-git/work-with-source-git/build-locally.md
TomasTomecek: several projects have hidden scripts which generate release archives, other projects utilize complex machinery (autotools) to create archives

so, for simple projects we could generate archives from git, for all gcc's, kernels, mysql's, cockpits it's much easier to download the existing archive

the other problem is that it would a major difference between production dist-git builds which use actual release archives

content/docs/source-git/work-with-source-git/build-locally.md
TomasTomecek: yup, sounds good!

content/docs/source-git/_index.md
TomasTomecek: no strong preference from my side, I just capitalized all

content/docs/source-git/_index.md
TomasTomecek: Maintainers need to know how patch files are going to be created out of the git commits and what packit exactly does to their spec file. Also, people who try source-git ask me this all the time.

content/docs/source-git/_index.md
TomasTomecek: I dropped the paragraph and instead elaborated in the point 1

content/docs/source-git/_index.md
TomasTomecek: Trouble = someone can change the spec file in the other place and suddenly builds would start failing while one would need to go and fix the spec which sits in a different repo. To me this feels like storing setup.py & setup.cfg in a different repo...

> If the spec file is stored some other place, the spec file can be changed without this change being reflected in the history of the source code. So tracing back a build to a given (source code, spec file) pair becomes problematic or would require additional tooling. Maybe this is the trouble you though of? slightly_smiling_face

exactly, that's why I don't understand the hatred for "spec file in upstream" scenario

content/docs/source-git/work-with-source-git/_index.md
TomasTomecek: I am going to drop that section for good, you're right that it doesn't feel right to have it here.

content/docs/source-git/work-with-source-git/_index.md
TomasTomecek: This is an old example which I did for Stream 8 and just moved it here, I'm gonna redo this section using a real upstream tag, that's a good point.

content/docs/source-git/work-with-source-git/_index.md
TomasTomecek: will add links, sounds like an overkill to explain pull request workflow here

content/docs/source-git/work-with-source-git/_index.md
TomasTomecek: yup, made it more accurate

content/docs/source-git/work-with-source-git/_index.md
TomasTomecek: We could certainly create an automation for this, that would be pretty awesome.

content/docs/source-git/work-with-source-git/_index.md
TomasTomecek: Since this is very technical, I'm personally fine with pointing to a PR (instead of not describing it at all) - are we able to cover this by an article? Would anyone read it or find it useful?

content/docs/source-git/_index.md
csomh: I see. Then I think this doc should explain these considerations. "Trouble" might not be self-explanatory enough to all readers :)

In general, my point is, that we should try to avoid using language that might be confusing to readers who didn't spend this much time studying the topic and we should go out of our way explaining the recommendations and decisions we took along the way. Writing "will save you a lot of trouble" doesn't meet any of this criteria.

csomh: :rocket:

content/docs/source-git/work-with-source-git/build-locally.md
csomh: Hm, I thought it's possible to configure how changelogs are generated, but that's a feature only when proposing dist-git updates with Packit Service.

content/docs/source-git/work-with-source-git/build-locally.md
csomh: https://github.com/packit/packit/issues/1096

content/docs/source-git/work-with-source-git/build-locally.md
csomh: Could we just use the `create-archive` action for the cases when the archive creation is not a trivial `git archive`?

TomasTomecek: yes, we could: I was actually thinking about something similar when I played with ruby: packit would download the archive by default and when create-archive is set, it would instead use those commands

content/docs/source-git/_index.md
TomasTomecek: I'll elaborate, good point.

csomh (approved):
Nice! :rocket:

PR #233, opened 2021-01-28
Run the 'deploy-pages' workflow on the main branch

csomh:
And publish the result to the 'main' page in the target repo.

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -3,7 +3,7 @@ name: Build and Deploy
 on:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   build:
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout master
-        uses: actions/checkout@v1
+      - name: Checkout branch
+        uses: actions/checkout@v2
         with:
           submodules: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-pages.yml
```diff
@@ -3,16 +3,16 @@ name: Build and Deploy
...
 jobs:
   build:
     if: github.repository == 'packit/packit.dev'
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout master
-        uses: actions/checkout@v1
+      - name: Checkout branch
+        uses: actions/checkout@v2
```
csomh, line 15: @jpopelka I [see](https://github.com/actions/checkout#checkout-v2) 'v2' is the recommended version for this action. Was there any reason for using v1 specifically when you've initially set up this workflow?

jpopelka: Because the [example @ the action's page](https://github.com/benmatselby/hugo-deploy-gh-pages#example) still says `v1`. Maybe even the action's author doesn't know about `v2`? :-)
Let's try your `v2` change and see.

jpopelka (approved):

jpopelka:
>  403 Resource not accessible by integration

I've already seen this a couple of times in this repo, wondering what/where might be the cause.

PR #244, opened 2021-03-01
Fix spelling mistakes

ochosi:
The correct perfect participle of "run" is "run", not "ran".


TomasTomecek (approved):
This is indeed correct, thank you for an English grammar lesson :)

PR #263, opened 2021-04-06
Fix paths in curl commands.

bcrocker15:
The curl commands as they were yielded HTML pages that said

    Page not found (404)

    With the message:

    Branch not found

    You have either entered a bad URL or the page has moved, removed, or otherwise rendered unavailable.
    Please use the main navigation menu to get (re)started.

The updated curl commands reflect the correct URLs; with the new curl
commands, it is possible to follow the example to completion.

Signed-off-by: Ben Crocker <bcrocker@redhat.com>


TomasTomecek (approved):
Very nice!

PR #284, opened 2021-05-24
fix,how-to-sg: add 'update-dist-git' to the shell output

TomasTomecek:
since the existing command would fail on the fact that no command was
specified


jpopelka (approved):

csomh:
I literally made this mistake every single time locally. Thanks for noticing and fixing it!

PR #288, opened 2021-05-27
Add documentation for patch_generation_patch_id_digits

csomh:
And also update the "Create source-git repo" guide to use this field.

### DIFF
--- a/content/docs/source-git/how-to-source-git.md
+++ b/content/docs/source-git/how-to-source-git.md
@@ -160,6 +160,7 @@ downstream_package_name: acl
 upstream_ref: v2.3.1
 patch_generation_ignore_paths:
   - .distro/
+patch_generation_patch_id_digits: 0
 sync_changelog: true
 synced_files:
   - src: .distro/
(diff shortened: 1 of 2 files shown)


content/docs/source-git/how-to-source-git.md
```diff
@@ -184,6 +185,10 @@ generating downstream patches, to be added in dist-git. As `.distro` is the
 place where the files related to packaging are stored, set the configuration
 to ignore this path, so that there are no patch files generated for it.
 
+`patch_generation_patch_id_digits` tells Packit how many digits should be used
+when adding `PatchN` tags to the spec-file. The patch in `acl` is in the form
+of `Patch1` (no leading zeros), so set this value to 0.
```
jpopelka, line 190: Hmm, I would say that 1 in `Patch1` is one digit and would set the value to 1.
So now it's not clear to me whether it's "number of digits" or "number of leading zeros".

csomh: Updated, please let me know if the new wording makes more sense.

This is a "minimum width in digits", where 0 means no minimum.

jpopelka (approved):

PR #292, opened 2021-06-04
Fix line breaks in the patch-metadata docs

csomh:
ce57d24 removed the space characters from the end of the lines, which
caused line-breaks to disappear.

I choose to restore the structure by turning the fields into paragraphs.
This should be less error-prone for the future. End-of-the-line
whitespaces can all to easily be removed by editors.


jpopelka (approved):

TomasTomecek (commented):
Looks good, thanks for fixing this!

(and agreed it's really weird that markdown chose such syntax for linebreaks)

PR #298, opened 2021-06-09
More master -> main

jpopelka:


content/docs/guide.md
```diff
@@ -43,7 +43,7 @@ On CentOS, some dependencies are in PowerTools repository, so you have to enable
 ### Via Fedora COPR
 
 You can also help us test the latest development snapshot by installing packit
-built from the master branch in Copr:
+built from the main branch in Copr:
 
     $ sudo dnf copr enable packit/packit-master
```
lachmanfrantisek, line 48: I am thinking about how we can get rid of this Copr project name.

Does Copr have some rename project option?  (And ideally, forward the requests for a while like on GitHub.) I remember we've discussed this but can't recall the outcome.

lachmanfrantisek (approved):
Thanks for looking at it.

content/docs/guide.md
jpopelka: I don't remember such a discussion and don't have answers to your questions, sorry. Thanks for the review!

PR #303, opened 2021-06-25
source-git update-dist-git: reflect recent code changes

TomasTomecek:
propose-downstream is no longer the command meant to be used with
source-git, we now have `packit source-git update-dist-git`: this
comment covers the changes

### DIFF
--- a/content/docs/source-git/work-with-source-git/propose-to-dist-git.md
+++ b/content/docs/source-git/work-with-source-git/propose-to-dist-git.md
@@ -8,33 +8,35 @@ weight: 3
 # Propose your source-git content to dist-git
 
 Once your changes are merged into a source-git repo, the final step is to
-propose those changes to dist-git and get a production build.
-
-For both, we have CLI commands in Packit which are meant to be run from the
-source-git's repo checkout.
-
-
-### Propose...
-
-    $ packit propose-downstream --dist-git-branch main
-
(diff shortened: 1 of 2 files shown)


content/docs/source-git/work-with-source-git/propose-to-dist-git.md
```diff
@@ -8,33 +8,35 @@ weight: 3
...
+environment. We advise you to inspect the changes done in your dist-git repo
+before pushing them out to be sure about them.
 
-Once the PR is up, you should wait for all the builds and tests to finish and
-succeed to be sure that all is good. When all checks shine with green, you are
-welcome to merge and celebrate for doing an update using source-git.
+If you are satisfied with the changes, put them in a new branch, push them out
+and create a merge request:
 
+    $ git checkout -B resolve-bz-1234567
```
jpopelka, line 33: I think new code/text should use
```suggestion
    $ git switch -C resolve-bz-1234567
```

jpopelka (approved):

content/docs/source-git/work-with-source-git/propose-to-dist-git.md
TomasTomecek: sounds good - I'm an oldskooler so it will take me a few years to stop using the checkout command :D

PR #316, opened 2021-07-28
source-git: Recommend 'patch_name' to merge commits

csomh:
Update the source-git metadata docs, to mark the 'squash_commits' key
deprecated and document that adjacent commits can be dumped in the same
patch file by setting the same 'patch_name' for them.

Related to packit/packit#1271.


jpopelka (approved):
:+1:

PR #381, opened 2022-01-31
Fix image reference for test runs in 2021

lbarcziova:


jpopelka (approved):
good catch

lachmanfrantisek (approved):
Thanks! Well spotted!

PR #382, opened 2022-02-01
Add create_sync_note option

majamassarini:
Related with packit/packit/issues#1356


content/docs/configuration.md
```diff
@@ -154,6 +154,14 @@ synced_files:
     mkpath: true
 ```
 
+#### create_sync_note
+
+(*bool*) When doing a new update in Fedora dist-git, if set to `true`,
+a *README.packit* file is created or updated with relevant sync information. 
+By default this option is `true`; when set to `false` the *README.packit* file will not be created or updated
+(if it was previously created it will not be updated and should be removed by hand).
```
csomh, line 162:
```suggestion
(*bool*) Create or update a *README.packit* file in dist-git when doing a new update. 
By default, this option is set to `true`. When set to `false`, *README.packit* is not created or updated.
(A previously created *README.packit* needs to be removed manually).
```

jpopelka (approved):

content/docs/configuration.md
csomh: Just changed the suggestion above to also tell what the content of *README.packit* is.

lachmanfrantisek (approved):

csomh (approved):

PR #389, opened 2022-02-15
Add blog-post for week 6

lachmanfrantisek:
Other changes are internal and does not need to be shown on blog.

### DIFF
--- a/content/posts/February-2022.md
+++ b/content/posts/February-2022.md
@@ -12,3 +12,8 @@ weight: 64
 - Packit-service now supports re-triggering against failed task using comment on
   pull request `/packit rebuild-failed` and similarly for testing farm: `/packit retest-failed`.
   ([packit-service#1303](https://github.com/packit/packit-service/pull/1303))
+
+## Week 6 (February 8th - February 14th)
+
+- Packit now correctly handles colons in git trailer values in source-git commits.
+  ([packit#1478](https://github.com/packit/packit/pull/1478))


content/posts/February-2022.md
```diff
@@ -12,3 +12,8 @@ weight: 64
 - Packit-service now supports re-triggering against failed task using comment on
   pull request `/packit rebuild-failed` and similarly for testing farm: `/packit retest-failed`.
   ([packit-service#1303](https://github.com/packit/packit-service/pull/1303))
+
+## Week 6 (February 8th - February 14th)
+
+- Packit now correctly handles colons in git trailer values in source-git commits.
+  ([packit#1478](https://github.com/packit/packit/pull/1478))
```
TomasTomecek, line 19: how about F36 and that we correctly support it right now? is it worth mentioning?

TomasTomecek (approved):
otherwise LGTM

content/posts/February-2022.md
lachmanfrantisek: I thought about that at first but don't think it's relevant to users. (And there was no changelog entry in the PR as well.) We basically haven't changed any behaviour.

TomasTomecek: From code PoV you're correct. But the fact that we'll restart pods and purge the cache will pick up F36 :)

I'm still LGTM, merge at your will.

PR #392, opened 2022-02-21
week 7

TomasTomecek:


content/posts/February-2022.md
```diff
@@ -17,3 +17,15 @@ weight: 64
 
 - Packit now correctly handles colons in git trailer values in source-git commits.
   ([packit#1478](https://github.com/packit/packit/pull/1478))
+
+## Week 7 (February 14th❤️ - February 18th)
+
+- You can now define `srpm_build_deps` as a top level key in your packit.yaml
+  to state which RPM dependencies are needed to create a SRPM. Next week (March
+  1st), Packit will take it into account and create your SRPM in Copr.
```
lbarcziova, line 25:
```suggestion
- You can now define `srpm_build_deps` as a top level key in your packit.yaml
  to state which RPM dependencies are needed to create a SRPM. Next week (March
  1st), Packit Service will take it into account and create your SRPM in Copr.
```
I wanted to add release notes only to packit-service PR since with packit change only, this has no effect on the user now. But with the explanation you provided, I am not against it, so I will leave it up to you.

lbarcziova (approved):
LGTM, just one note

content/posts/February-2022.md
TomasTomecek: I'm torn right now. Both options make sense.

After reading your comment, you're correct this feature is not actionable to users yet. They can only test it on stg once your PR is merged. Let me rephrase and we'll see how it looks.

TomasTomecek: Okay, I'm scraping it. People would need packit CLI from Copr since the config change didn't make it to the latest release. Let's leave this to the next week. Thank you for your honest feedback!

PR #393, opened 2022-02-22
Document 'srpm_build_deps' configuration key

lbarcziova:


jpopelka (approved):

content/docs/configuration.md
```diff
@@ -309,6 +309,12 @@ sources:
 corresponding basename exists in the spec-file, the source will be downloaded from the `url` found in the configuration
 instead of the location defined in the spec-file.
 
+
+#### srpm_build_deps
+(*list of strings*) A list of RPM dependencies that are needed for your actions to be run when building SRPM. 
+This key is used in Packit Service as an indicator to build your SRPMs in Copr and 
+the dependencies are then installed into the Copr build environment each time the build is triggered.
```
lachmanfrantisek, line 316: Can we add a note that we want to switch to Copr for all packages in the future?

Something like this?

```suggestion
#### srpm_build_deps
(*list of strings*) A list of RPM dependencies that are needed for your actions to be run when building SRPM. 
For now, this key is used in Packit Service as an indicator to build your SRPMs in Copr and 
the dependencies are then installed into the Copr build environment each time the build is triggered.
Be aware that this is a preparation phase and we want to build all the packages in Copr by default.
```

PR #399, opened 2022-02-28
Update `tmt` examples to use the current spec

psss:
Update the obsoleted key names.
Remove virtual hierarchy to prevent confusion.
Cleanup up tabs causing weird indent.


mfocko (approved):
lgtm, thanks for the update

TomasTomecek (approved):
I really like the simplifications, thanks!

PR #416, opened 2022-03-21
Users need to be careful about syntax when using `copy_upstream_release_description`

lachmanfrantisek:
Relates to: https://github.com/packit/packit/issues/1521


TomasTomecek (approved):
thank you!

PR #424, opened 2022-04-01
Document the newly added Fedora-aliases

csomh:
Related to packit/packit#1546.

### DIFF
--- a/content/docs/configuration.md
+++ b/content/docs/configuration.md
@@ -438,6 +438,36 @@ jobs:
 In this example, the job `some-job` would override `specfile_path` to
 `somewhere/else/package.spec` instead of using `./package.spec`.
 
+#### Aliases
+
+To not need to change the config file when the new system version is released,
+Packit provides multiple aliases to reference a subset of the active
+Fedora Linux releases:
+* `fedora-all` - all active releases, which includes released and branched
+  versions and Rawhide (e.g. `fedora-34`, `fedora-35`, `fedora-36`,
+  `fedora-rawhide`).
+* `fedora-stable` — the current (two or three) released and supported
+  versions (e.g. `fedora-34`, `fedora-35`).
+* `fedora-development` — development versions of Fedora; the branched
+  version is used only when available (e.g. `fedora-36`, `fedora-rawhide`)
+* `fedora-latest` — the last versioned Fedora (not a Rawhide), regardless if
+  it's released or still under development.
+* `fedora-latest-stable` — the latest released Fedora version (e.g.
+  `fedora-35`).
+* `fedora-branched` — all branched releases, that is: everything, except
+Rawhide (e.g. `fedora-34`, `fedora-35`, `fedora-36`).
+
+Additionally, `epel-all` can be used as an alias for the current active
+[EPEL](https://docs.fedoraproject.org/en-US/epel/) versions (e.g. `epel-7`,
+`epel-8`, `epel-9`)
+
+The aliases above can be used both to specify targets when [building in
+Copr](#copr_build) or [running tests](/testing-farm/), and to reference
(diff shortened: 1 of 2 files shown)


content/docs/configuration.md
```diff
@@ -523,16 +523,27 @@ fedora-32-armhfp
...
-  * `epel-all` — all available versions of epel (currently `el-6`, `epel-7` and `epel-8`)
-
-  The information about releases is get from Bodhi and because of the cache and
-  required availability on Copr, it might take a while to get the newest state.
+* It's also possible to use aliases, to reference a subset of the active
+  Fedora Linux releases:
+  * `fedora-all` - all active releases, which includes released and branched
+    versions and Rawhide (e.g. `fedora-34`, `fedora-35`, `fedora-36`,
+    `fedora-rawhide`).
+  * `fedora-stable` — the current (two) released and supported versions (e.g.
```
lachmanfrantisek, line 531:
```suggestion
  * `fedora-stable` — the current (two or three) released and supported versions (e.g.
```

or

```suggestion
  * `fedora-stable` — the current (usually two) released and supported versions (e.g.
```

lachmanfrantisek (commented):
Thanks!

Would it be possible to move it to a dedicated section?
* it's not only related to Copr builds
* it can be linked from elsewhere

csomh:
> Thanks!
> 
> Would it be possible to move it to a dedicated section?
> 
> * it's not only related to Copr builds
> * it can be linked from elsewhere

I'm not sure where could I move this... Any suggestions?

Aliases are related to Copr build targets, so I think this is a good place to discuss them. And the section is referenced when discussing [test targets](https://github.com/packit/packit.dev/blob/56d52ffa0c7c3233a6c09bcbae082d5b66fca144/content/docs/testing-farm.md?plain=1#L40).

lachmanfrantisek:
> I'm not sure where could I move this... Any suggestions?
>
> Aliases are related to Copr build targets, so I think this is a good place to discuss them. And the section is referenced when discussing [test targets](https://github.com/packit/packit.dev/blob/56d52ffa0c7c3233a6c09bcbae082d5b66fca144/content/docs/testing-farm.md?plain=1#L40).

Maybe as another second-level header in the `docs/configuration`? It can actually be used across the whole config and also in CLI arguments -- we can convert the alias to Copr targets, dist-git branches, ...

csomh:
> > I'm not sure where could I move this... Any suggestions?
> > Aliases are related to Copr build targets, so I think this is a good place to discuss them. And the section is referenced when discussing [test targets](https://github.com/packit/packit.dev/blob/56d52ffa0c7c3233a6c09bcbae082d5b66fca144/content/docs/testing-farm.md?plain=1#L40).
> 
> Maybe as another second-level header in the `docs/configuration`? It can actually be used across the whole config and also in CLI arguments -- we can convert the alias to Copr targets, dist-git branches, ...

I'm not sure I understand how this should look like. Can you please provide a concrete suggestion/git commit?

lachmanfrantisek:
I can't suggest outside of your changes so this is what I meant:

```
:

In this example, the job `some-job` would override `specfile_path` to
`somewhere/else/package.spec` instead of using `./package.spec`.

#### Aliases

E.G. HERE

#### Supported jobs

:
```

The goal is to have a header for that part so we can use an anchor link in the job descriptions. E.g.:

```
* **dist_git_branches** - a (list of) branch(es) in dist-git where packit should work (defaults to `main`).
  You can also use [aliases](#aliases)
```


Also, this link could be used to link to aliases:
https://packit.dev/docs/configuration/#aliases

lachmanfrantisek:
recheck

TomasTomecek (approved):

lachmanfrantisek (approved):

PR #427, opened 2022-04-04
Update CLI/source-git/update-source-git

jpopelka:
See packit/packit#1547


csomh (approved):
The commit message could be made self-contained, by describing that this change is about making the revision range argument optional for 'source-git update-source-git'.

PR #430, opened 2022-04-11
Add downstream_branch_name documentation

majamassarini:


content/docs/configuration.md
```diff
@@ -183,6 +183,11 @@ pattern with `branches/`, e.g. for a branch matching `*-release` set to `branche
 (*string*) Name of the RPM package in Fedora, defaults to the name of the
 Github repository.
 
+#### downstream_branch_name
+
+(*string*) Name of the branch in dist-git which corresponds to the current source-git branch,
+which is meant to be used in source-git projects only.
```
csomh, line 189:
```suggestion
(*string*) Name of the branch in dist-git which corresponds to the current source-git branch. Defaults to the name of the current source-git branch. This is meant to be used in source-git projects only.
```

csomh (commented):
The usual nitpicks :slightly_smiling_face:

csomh (approved):

lachmanfrantisek (commented):
I know I am really late for the party, but isn't this a bit confusing with the `dist_git_branches` metadata key?

Also in the sense that we want to move the metadata keys directly to the jobs/top-level... So we will have both `downstream_branch_name` and `dist_git_branches`.

(Sorry if you've already discussed this...)

csomh:
I'm not sure, if this aspect was discussed, to be honest. But we can revisit this when `dist_git_branches` becomes top-level. Until then `downstream_branch_name` is the key hardly is looking at. (Note: `source-git.yaml` doesn't have `jobs`.)

lachmanfrantisek:
> I'm not sure, if this aspect was discussed, to be honest.

Actually, it was. At least on the arch meeting on Mar 10, 2022 what resulted in this issue: https://github.com/packit/packit/issues/1515

> But we can revisit this when dist_git_branches becomes top-level. Until then downstream_branch_name is the key hardly is looking at. 

Solving this after that change is too late I think --  we would need to preserve the old naming.

Is it a big issue to change that in Hardly now? I can help with that if you want. Is the main issue in the change itself or the difference between those two that one is a list and one a single value (what we used historically for `propose-downstream` as well)?

> (Note: source-git.yaml doesn't have jobs.)

OK, that's a good point, but still can make some confusion:
* The key is mentioned in the documentation together with other Packit options.
* The key matches the schema => there won't be an error when the user chooses a bad one.
* Once `dist_git_branches` are moved to the top-level, we can use that as a default for CLI instead of searching through the jobs (and get rid of this not-ideal solution). So, strictly CLI users can avoid the `jobs` key in the same way as source-git users do.

jpopelka:
I've just added it to the arch meeting agenda we have tomorrow morning.

PR #432, opened 2022-04-13
Update source-git docs

csomh:
Revisit the design document and workflow pages, move the source-git documentation to the top of the hierarchy, remove redundancies and outdated content.

PR #439, opened 2022-04-26
fix: Apply source-git doc suggestions from Tomas

csomh:
In response to the comments in #432 which were made after merging it.


TomasTomecek (approved):
Thank you for getting back to this.

PR #441, opened 2022-04-28
content/docs/faq.md: Use %autorelease instead of deprecated %autorel

ueno:
The `%autorel` macro was renamed to `%autorelease` in rpmautospec 0.2:
https://docs.pagure.org/fedora-infra.rpmautospec/autorelease.html

Also make sure to include `%{?dist}` if `%autorelease` is not defined.

Signed-off-by: Daiki Ueno <dueno@redhat.com>


jpopelka (approved):
thanks!

PR #449, opened 2022-05-05
Aliases provided by packit should be used for targets - not for branch

nikromen:
If you feel that documentation for `branch` metadata is too short - your suggestion is welcomed :innocent:

### DIFF
--- a/content/docs/configuration.md
+++ b/content/docs/configuration.md
@@ -605,13 +605,13 @@ Supported triggers:
 Required metadata:
 
 * **branch** -- the name of the branch we want to build for when using **commit** trigger.
-  You can also use the [aliases provided by Packit](#aliases)
-  to not need to change the config file when the new system version is released.
 
 Optional metadata:
 
 * **targets** -- a (list of) targets we want to build for,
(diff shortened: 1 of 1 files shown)


content/docs/configuration.md
```diff
@@ -605,13 +605,13 @@ Supported triggers:
...
 * **branch** -- the name of the branch we want to build for when using **commit** trigger.
-  You can also use the [aliases provided by Packit](#aliases)
-  to not need to change the config file when the new system version is released.
 
 Optional metadata:
 
 * **targets** -- a (list of) targets we want to build for,
   list of supported targets can be listed using with `koji list-targets`
+  You can also use the [aliases provided by Packit](#aliases)
+  to not need to change the config file when the new system version is released.
```
lachmanfrantisek, line 614:
```suggestion
* **targets** -- a (list of) targets we want to build for,
  list of supported targets can be listed using with `koji list-targets`.
  You can also use the [aliases provided by Packit](#aliases)
  to not need to change the config file when the new system version is released.
```

lachmanfrantisek (approved):
Thanks!

PR #456, opened 2022-05-12
Move release_suffix out of the way of files_to_sync examples

csomh:
Signed-off-by: Hunor Csomortáni <csomh@redhat.com>

PR #457, opened 2022-05-12
Release suffix at the top-level

mfocko:
I have also rebased against #456, so I'm not sure if we want to merge them in succession or just this one.

Closes #456


TomasTomecek (approved):
Can you please also document the default %release Packit sets and resolve https://github.com/packit/packit/issues/1419?

otherwise LGTM

Issue #429, opened 2022-04-11
Navigation: blog posts ordering is confusing


TomasTomecek:
```
[Blog Posts]()
- [Building SRPMs in Copr](https://packit.dev/posts/copr-srpms/)
- [March 2022](https://packit.dev/posts/march-2022/)
- [February 2022](https://packit.dev/posts/february-2022/)
- [2021 for Packit](https://packit.dev/posts/2021-features/)
- [January 2022](https://packit.dev/posts/january-2022/)
- [2021 in Numbers](https://packit.dev/posts/2021-in-numbers/)
- [December 2021](https://packit.dev/posts/december-2021/)
- [November 2021](https://packit.dev/posts/november-2021/)
- [October 2021](https://packit.dev/posts/october-2021/)
- [September 2021](https://packit.dev/posts/september-2021/)
- [August 2021](https://packit.dev/posts/august-2021/)
- [July 2021](https://packit.dev/posts/july-2021/)
- [June 2021](https://packit.dev/posts/june-2021/)
- [May 2021](https://packit.dev/posts/may-2021/)
- [April 2022](https://packit.dev/posts/april-2022/)
- [April 2021](https://packit.dev/posts/april-2021/)
- [March 2021](https://packit.dev/posts/march-2021/)
- [February 2021](https://packit.dev/posts/february-2021/)
- [January 2021](https://packit.dev/posts/january-2021/)
- [Working on the next major RHEL release, in your upstream repo](https://packit.dev/posts/fedora-eln/)
- [2020](https://packit.dev/posts/2020/)
- [2019](https://packit.dev/posts/2019/)
```

I'd do several improvements here:
- put weekly blog posts into a subcategory
- order by date (newest first)
- Think about 2021, should it be compressed?

Example, please suggest a better structure:
```
Blog posts
- [Building SRPMs in Copr](https://packit.dev/posts/copr-srpms/)
- [2021 in Numbers](https://packit.dev/posts/2021-in-numbers/)
- [2021 for Packit](https://packit.dev/posts/2021-features/)
- [Working on the next major RHEL release, in your upstream repo](https://packit.dev/posts/fedora-eln/)
- Weekly updates
  - April 2022
  - March 2022
  - February 2022
  - January 2022
  - 2021
  - 2020
  - 2019
```

You can also get inspired by mock's releases list: https://rpm-software-management.github.io/mock/#release-notes

PR #458, opened 2022-05-16
Blogposts improvements

FrNecas:
Fixes: #429


FrNecas:
Navigation after the updates:

screenshot_20220516_145550

TomasTomecek (approved):
Well, that's exactly what I was asking for :) so LGTM

I know that when we refined this card, folks had comments about the squashing and structure, so let's see what other reviewers will say

lbarcziova (approved):
:+1:

majamassarini:
I run the Hugo server locally without before looking at this page. 
I should say that I was totally blind and could't find the Weekly Updates at first sight. After I saw your image in the comments here I could finally find them.

If we want the Weekly Updates to be more visible maybe we should raise them one level? Otherwise if we want them not to make too much noise is fine like it is now.

FrNecas:
Renamed Weekly Updates to Weekly Status and moved it to the top as agreed on arch. I am trying to look into RSS in hugo right now

FrNecas:
RSS feed with hugo works out of the box: https://packit.dev/posts/index.xml . With the new structure, also https://packit.dev/posts/weekly/index.xml will work. I am merging this since it was approved and no significant changes were made by the last push (just things we agreed on at arch) :)

PR #460, opened 2022-05-17
Remove rpmautospec link from the heading

FrNecas:
Hugo does not seem to like links in the headings since it creates a
nested link resulting in weird indentation/line breaks. Keep the link to
rpmautospec only in the paragraph.

See the attached screenshots for before and after (in the before screenshot, the nav in the right column is also somewhat broken)

Before:

screenshot_20220517_150405

After:

screenshot_20220517_150539


jpopelka (approved):
Hmm, strange that nobody noticed so far. Thanks

PR #477, opened 2022-06-03
autorelease usage suggestion fallback case seems wrong

t184256:
https://packit.dev/docs/faq/#does-packit-work-with-rpmautospec says the following:

> `Release:  %{?autorelease}%{!?autorelease:1{?dist}}`
>
> This construct uses `autorelease` macro if it’s defined, and if it’s not, it sets release to `1{?dist}`.

I'm afraid this statement is technically correct, and, if `autorelease` macro happens to be undefined, the expression evaluates to a literal `1{?dist}` =)

It was meant to be `%{?autorelease}%{!?autorelease:1%{?dist}}` with a `%` in `%{?dist}`, right?

### DIFF
--- a/content/docs/faq.md
+++ b/content/docs/faq.md
@@ -140,9 +140,9 @@ rpmautospec utilizes two RPM macros:
 
 If you want your upstream spec file to also work well when `rpmautospec-rpm-macros` is not installed, set `Release` to this:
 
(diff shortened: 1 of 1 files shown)


jpopelka (approved):
Yup, thanks for the fix!

PR #484, opened 2022-06-14
Copr defaults to enable_net=False

praiskup:
https://pagure.io/copr/copr/blob/922e97d485718653a526ad1b3203597a226f9278/f/frontend/coprs_frontend/coprs/forms.py#_577-579


TomasTomecek (approved):
Thanks!

TomasTomecek (approved):
even better :)

PR #493, opened 2022-07-11
Create blog post for weeks 26–27, 2022

csomh:


lbarcziova (commented):
since we skipped the moving of the branches last week, we should probably include the updates for the previous week here, right?

csomh:
> since we skipped the moving of the branches last week, we should probably include the updates for the previous week here, right?

Right. Updated.

PR #508, opened 2022-07-29
guide: improve section 'check that Packit works'

TomasTomecek:
More details on what's happening in the background, especially the 'merge
PR' part.

Related: https://github.com/packit/packit/issues/1654

### DIFF
--- a/content/docs/guide.md
+++ b/content/docs/guide.md
@@ -356,11 +356,18 @@ since we try to use all the latest features.
 
 ## Check that Packit works
 
-With the configuration above, Packit is able only to build your changes and provide them as RPMs.
-Once you create a pull request, Packit builds the code from the
-pull request in Copr repository and reports back Copr build status to the pull request.
+With the configuration above, Packit will build your changes and provide them as RPMs.
+Once you create a pull request, Packit will:
+1. Check out the PR.
+2. Merge it to the target branch (this can be configured by the
+   [`merge_pr_in_ci`](https://packit.dev/docs/configuration/#merge_pr_in_ci)
+   configuration option).
+3. Submit a build using the code that results from the merge in the Copr build
+   system [in Packit's
+   namespace](https://copr.fedorainfracloud.org/coprs/packit/).
+4. Report results about the Copr build status to the pull request.
 
-An example of Packit pull request:
+An example of Packit's checks in a pull request:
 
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
Thanks!

content/docs/guide.md
```diff
@@ -356,11 +356,17 @@ since we try to use all the latest features.
...
 
-With the configuration above, Packit is able only to build your changes and provide them as RPMs.
-Once you create a pull request, Packit builds the code from the
-pull request in Copr repository and reports back Copr build status to the pull request.
-
-An example of Packit pull request:
+With the configuration above, Packit will build your changes and provide them as RPMs.
+Once you create a pull request, Packit will:
+1. Check out the PR.
+2. Merge it with the main branch (this can be configured by the
```
pcahyna, line 362: I suppose you mean the target branch of the PR, whatever its name is ("main branch" could be misunderstood as the branch whose name is "main", or the default branch of the repo, like `main`, `master` `devel`, whatever).
The difference shows up when you open a PR against another branch than the one you usually merge to.

content/docs/guide.md
```diff
@@ -356,11 +356,17 @@ since we try to use all the latest features.
...
+   [`merge_pr_in_ci`](https://packit.dev/docs/configuration/#merge_pr_in_ci)
+   configuration option).
+3. Submit a build with the code from the pull request in Copr [in its
```
pcahyna, line 365: Is this true? My understanding is that it will submit a build with the code resulting from the merge, not the code in the PR (that's probably what you meant, but the wording is misleading).

content/docs/guide.md
lachmanfrantisek: Yes, Pavel is right. We merge to the target branch.

content/docs/guide.md
TomasTomecek: @pcahyna thank you! both of your suggestions definitely make the explanation more clear

content/docs/guide.md
pcahyna: For comparison, here's how Zuul documents its procedure (it uses "target branches"). https://zuul-ci.org/docs/zuul/5.2.2/job-content.html#git-repositories

pcahyna:
thanks, looks good now

PR #529, opened 2022-09-19
Add blogpost for week 37

FrNecas:
Fixes https://github.com/packit/weekly-roles/issues/31

### DIFF
--- a/content/posts/weekly/September-2022.md
+++ b/content/posts/weekly/September-2022.md
@@ -22,3 +22,8 @@ weight: 58
   packit now correctly handles pagination of API results. This resolves an issue that caused
   Fedora 37 not to be pointed to from any alias.
   ([packit#1704](https://github.com/packit/packit/pull/1704))
+
+## Week 37 (September 13th – September 19th)
+
+- Added support for filenames specified in source URL fragments, for example: `https://example.com/foo/1.0/download.cgi#/%{name}-%{version}.tar.gz` ([specfile#100](https://github.com/packit/specfile/pull/100))
+- Some more underlying improvements to our libraries and deployment.


FrNecas:
Whoops, didn't have pre-commit setup, will fix this

content/posts/weekly/September-2022.md
```diff
@@ -22,3 +22,8 @@ weight: 58
   packit now correctly handles pagination of API results. This resolves an issue that caused
   Fedora 37 not to be pointed to from any alias.
   ([packit#1704](https://github.com/packit/packit/pull/1704))
+
+## Week 37 (September 13th – September 19th)
+
+- Using the method `users_with_write_access` you can generate the set of users that have write access to the project and the method `has_write_access(user)` you can find out if the user has write access to the project. ([ogr#742](https://github.com/packit/ogr/pull/742))
+- Added support for filenames specified in source URL fragments, for example: `https://example.com/foo/1.0/download.cgi#/%{name}-%{version}.tar.gz` ([specfile#100](https://github.com/packit/specfile/pull/100))
```
lachmanfrantisek, line 29: I would not mention these in the weekly blog post -- these are more about OGR and Specfile respectively.

You can provide a general sentence about underlying improvements...

content/posts/weekly/September-2022.md
```diff
@@ -22,3 +22,8 @@ weight: 58
   packit now correctly handles pagination of API results. This resolves an issue that caused
   Fedora 37 not to be pointed to from any alias.
   ([packit#1704](https://github.com/packit/packit/pull/1704))
+
+## Week 37 (September 13th – September 19th)
+
+- Using the method `users_with_write_access` you can generate the set of users that have write access to the project and the method `has_write_access(user)` you can find out if the user has write access to the project. ([ogr#742](https://github.com/packit/ogr/pull/742))
```
TomasTomecek, line 28:
```suggestion
- With our library ogr you can now use the method `users_with_write_access` to generate a set of users that have write access to a project. Method `has_write_access(user)` can find out if a user has write access to the project. ([ogr#742](https://github.com/packit/ogr/pull/742))
```
Was really confused until I realized it's an ogr change. Let's make this obvious from the start.

TomasTomecek (changes requested):
The second item is nice and clear.

content/posts/weekly/September-2022.md
FrNecas: I too was wondering about this because I remember making the changes to `move-stable` to use `changelog.py` but only on packit/packit-service but https://github.com/packit/deployment/pull/364 added also specfile and ogr so I thought that there was a consensus to also include those.

FrNecas: I think that in blog posts we should really focus only on user-facing changes so ignore the libraries (unless it fixes an important bug for our users). I will remove ogr and specfile from the produced blogpost and drop the line about ogr from this week since it's very specific. As Tomas mentioned, the specfile one may be useful

nforro: I agree with Tomáš in mentioning the specfile change, it manifested as a bug (though harmless) in packit after all.

lbarcziova (approved):

FrNecas:
Updated. Thinking about it a bit more, it makes sense to keep ogr/specfile in blogpost generation (I don't want to get these manually in case they are relevant - e.g. bugfixes like the specfile one). But I am going to update the service guru card to be a bit more explicit about this.

TomasTomecek (approved):

FrNecas:
Prod seems to be running just fine, merging the blogpost :)

PR #563, opened 2022-11-24
Fix detached links @ control-patch-generation page

webknjaz:
The syntax was previously broken because these definitions weren't all on separate lines.


webknjaz:
cc @TomasTomecek

I accidentally got on the https://packit.dev/source-git/work-with-source-git/control-patch-generation/#no-prefix page from Google and saw something broken at the very bottom of the page.

csomh:
Thanks @webknjaz!

csomh (approved):

PR #578, opened 2022-12-30
correct syntax of `koji build` link

evgeni:


TomasTomecek (approved):
thanks, very nice catch!

PR #604, opened 2023-02-16
rpminspect: run using c9s configuration

### LINKED ISSUE #601: Document how to run rpminspect using the CentOS Stream configuration
We already document how to run rpminspect, let's extend that documentation

https://github.com/packit/packit.dev/blob/bbd3ae77f346acb4e6ee968c41ce23670ae20a4e/content/docs/testing-farm.md#rpminspect

Profile for CentOS Stream: https://gitlab.com/redhat/centos-stream/ci-cd/rpminspect-data-centos 

Ideally, find someone who'd be wiling to try this out.

It may be worth starting a new document specifically for RHEL/Stream maintainers.

TomasTomecek:
This should help everyone who maintains a package in CentOS Stream 9 to
get rpminspect feedback upstream using the same configuration as when
contributing to c9s.

Fixes #601

### DIFF
--- a/content/docs/testing-farm.md
+++ b/content/docs/testing-farm.md
@@ -289,7 +289,7 @@ execute:
 
 #### rpminspect
 
-[`rpminspect`](https://github.com/rpminspect/rpminspect) can analyze your SRPMs
+[`rpminspect`](https://github.com/rpminspect/rpminspect) can analyze your packages
 and give you information related to licensing, metadata, manpages, desktop app
 metadata, file ownership & permissions and much much more.
 
@@ -297,7 +297,7 @@ Here's a tmt plan you can use to have rpminspect invoked on SRPMs produced by Pa
 
 ```yaml
 summary:
-    Fetch SRPM and check it with rpminspect
+    Check rpm files with rpminspect
 discover:
     how: shell
     tests:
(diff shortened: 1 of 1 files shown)


mfocko (approved):

content/docs/testing-farm.md
```diff
@@ -315,6 +315,27 @@ execute:
...
+    how: shell
+    tests:
+      - name: rpminspect
+        test: rpminspect-centos -v -t VERIFY --profile=centos-stream-9-devel /tmp/*.src.rpm
+prepare:
+  - name: packages
+    how: install
+    package:
+    - rpminspect
+    - rpminspect-data-fedora
```
msrb, line 332: I think you probably meant [rpminspect-data-centos](https://copr.fedorainfracloud.org/coprs/dcantrell/rpminspect/packages/).

content/docs/testing-farm.md
```diff
@@ -315,6 +315,27 @@ execute:
...
 
+You can run rpminspect also using the CentOS Stream configuration. This should prepare you before opening CentOS Stream dist-git MRs.
+```yaml
+summary:
+    Fetch SRPM and check it with rpminspect
+discover:
+    how: shell
+    tests:
+      - name: rpminspect
+        test: rpminspect-centos -v -t VERIFY --profile=centos-stream-9-devel /tmp/*.src.rpm
```
msrb, line 326: Any reason why to run rpminspect only on the SRPM? Binary packages are well supported as well (it's actually where rpminspect truly shines).

msrb:
I think this is a good start and I am surprised how easy it looks.

One thing to keep in mind is that rpminspect in Fedora/EPEL is usually old and buggy. The tool is under active development and thus CentOS Stream uses the latest builds from David's COPR: https://copr.fedorainfracloud.org/coprs/dcantrell/rpminspect/

Another thing is that rpminspect compares the new build with the previous build. It can run on just the new build, but then inspections like `abidiff` won't run as there is no previous build to compare the ABI to.

And the last thing is that maintainers in CentOS Stream can create their own `rpminspect.yaml` files and store them in dist-git. This local component-specific configuration tweaks the behavior of rpminspect.

But packit doesn't have to support all that.

content/docs/testing-farm.md
TomasTomecek: thank you for spotting this

content/docs/testing-farm.md
TomasTomecek: good point! quite honestly, I wasn't aware of that

trying to figure out with TF/tmt if those are already available in the environment because there is no easy way to download them (by easy I mean I'd need to write an ugly shell script)

TomasTomecek:
Trying the proposal here: https://github.com/packit/ogr/pull/770

Bug report: https://gitlab.com/testing-farm/general/-/issues/65

TomasTomecek:
I will create a followup card regarding the runs on binary RPMs as we need this to be implemented in testing farm: https://issues.redhat.com/browse/TFT-1574

PR #613, opened 2023-02-28
Mention that we have our identity on salsa.debian.org

lachmanfrantisek:


lachmanfrantisek:
Hmm...;( Looks like we need to fix the `packit prepare-sources` as well.)

lachmanfrantisek:
The easiest way, for now, is to hardcode such hostnames: https://github.com/packit/ogr/pull/775

(Doing a test call to get that automatically would require more work and would result in extra API queries.)

TomasTomecek (approved):

lbarcziova (approved):

PR #619, opened 2023-03-15
Add workflow image to the development docs

lachmanfrantisek:
I've already created this so why not placing it somewhere so we can share it with someone easily.

image
image


TomasTomecek (approved):
I love how playful it is!

PR #627, opened 2023-03-28
Document using `copr` GSSAPI

LecrisUT:
IMO this should be the default configuration over the API tokens.


jpopelka (approved):
Works for me, thanks!

lbarcziova (approved):
thanks!

PR #654, opened 2023-05-22
Tree map usage graphs

lachmanfrantisek:
This is a really basic static version of the charts. There is a script that you can use to regenerate the charts. In the future, we will (ideally) reuse them in the dashboard instead.

Screenshot from 2023-05-22 17-12-34


majamassarini (approved):
Nice!

TomasTomecek (approved):
sweet, I love you added all jobs!

lbarcziova (commented):
nice! :rocket:

how do we want to use this currently? Link it from dashboard? Or add a menu item linking the new page here in packit.dev?

> In the future, we will (ideally) reuse them in the dashboard instead.

agreed, we should consider merging/replacing some parts of the current https://dashboard.packit.dev/usage to not duplicate the info

lachmanfrantisek:
> how do we want to use this currently? Link it from dashboard? Or add a menu item linking the new page here in packit.dev?

For now, I wouldn't treat that seriously and don't want to link that from anywhere (maybe just send the link if someone is interested).

The information is really similar to what we already have on the dashboard's usage page so we need to think about that more properly...

PR #659, opened 2023-05-30
Add doc about manual_trigger option for testing farm

Frawless:


Frawless:
cc @lbarcziova

lbarcziova:
Hi, thanks for the contribtion!

As described in https://github.com/packit/packit-service/issues/1806#issuecomment-1569516694, there is still an issue regarding the dependant jobs, so I would wait with the documentation until this works for every use case.

content/docs/testing-farm.md
```diff
@@ -108,6 +118,7 @@ You can verify that the option is processed correctly using a [YAML parser](http
   when you have multiple `tests` jobs with different configuration. For
   example if you set this to `e2e-tests`, then a check run for Rawhide would be
   named `testing-farm:fedora-rawhide-x86_64:e2e-tests`.
+* **manual_trigger** - Whether to trigger Testing Farm jobs only via pull request comment or not (default to false). Trigger phrase is `/packit test` (`/packit-stg test` for stage).
```
psss, line 121: How do you specify the job which should be triggered? Should we include an example for that as well?

lbarcziova: as @Frawless wrote, currently, all test jobs will be triggered. 

The follow-up issue that would cover the functionality of specifying the job: https://github.com/packit/packit-service/issues/1754#issuecomment-1522949476

psss: Ah, I see, so this is the first part. Thanks for clarification!

content/docs/testing-farm.md
```diff
@@ -60,6 +60,16 @@ jobs:
...
 ```
 
+If you want to trigger tests via pull request comment and not by every new commit into the pull request, the test job needs to include `manual_trigger` (described below) option in the job configuration:
+```yaml
+  jobs:
+  - job: tests
+    trigger: pull_request
+    targets:
+    - fedora-all
+    manual_trigger: true
```
lbarcziova, line 70:
```suggestion
    skip_build: true
    manual_trigger: true
```

just for clarity, since without `skip_build` enabled, the build job needs to be configured as well

content/docs/testing-farm.md
lbarcziova, line 121:
```suggestion
* **manual_trigger** - Whether to trigger Testing Farm jobs only manually (via pull request comment `/packit test`  (`/packit-stg test` for staging instance) or rerunning the check in the GitHub UI) or not (defaults to false). 
```

lbarcziova (commented):
thanks! This should now work in general (also for builds and propose-downstream), so it should be mentioned in the general docs as well. If you do not have time to do that, I can cover that part :)

Frawless:
> thanks! This should now work in general (also for builds and propose-downstream), so it should be mentioned in the general docs as well. If you do not have time to do that, I can cover that part :)

Do you mean here https://github.com/packit/packit.dev/blob/main/content/docs/Configuration/examples.md ?

content/docs/testing-farm.md
```diff
@@ -157,6 +169,11 @@ for the targets configured in the repository with the "main" pull request.
 This is a new feature, so the behaviour may be adjusted in the future. 
 Please reach out back to us for help or with your suggestions.
 
+## Running tests with specific identifier
+It is possible to run a specific job via `/packit test` command. 
+User just need to specify argument `--identifier <job_identifier>` and Packit will trigger only job with this identifier.
+The whole commnad should looks like this: `/packit test --identifier my-job-id`
```
lbarcziova, line 175:
```suggestion
The user just needs to specify the argument `--identifier <job_identifier>` and Packit will trigger only the job with this identifier.
The whole command should look like this: `/packit test --identifier my-job-id`
```

lbarcziova (commented):
otherwise LGTM, thanks!

content/docs/testing-farm.md
```diff
@@ -108,6 +145,8 @@ You can verify that the option is processed correctly using a [YAML parser](http
   when you have multiple `tests` jobs with different configuration. For
   example if you set this to `e2e-tests`, then a check run for Rawhide would be
   named `testing-farm:fedora-rawhide-x86_64:e2e-tests`.
+* **manual_trigger** - Whether to trigger Testing Farm jobs only manually (via pull request comment `/packit test`  (`/packit-stg test` for staging instance) or rerunning the check in the GitHub UI) or not (defaults to false). 
+* **labels** - List of labels that groups several jobs together. Users then use them when manully trigger the jobs like `/packit test --labels regression,upgrade`. The labels should be in the format of comma-separated string.
```
lbarcziova, line 149:
```suggestion
* **labels** - List of labels that group several jobs together. Users then use them when manually triggering the jobs like `/packit test --labels regression,upgrade`.
```

lbarcziova (commented):
I would probably move the note about comma-separated format into the section about running the command (see the suggestions), WDYT?

content/docs/testing-farm.md
```diff
@@ -157,6 +196,16 @@ for the targets configured in the repository with the "main" pull request.
 This is a new feature, so the behaviour may be adjusted in the future. 
 Please reach out back to us for help or with your suggestions.
 
+## Running tests with specific identifier
```
lbarcziova, line 199:
```suggestion
## Running tests with a specific identifier
```

content/docs/testing-farm.md
```diff
@@ -157,6 +196,16 @@ for the targets configured in the repository with the "main" pull request.
...
+It is possible to run a specific job via `/packit test` command. 
+The user just needs to specify the argument `--identifier <job_identifier>` and Packit will trigger only the job with this identifier.
+The whole command should look like this: `/packit test --identifier my-job-id`
+
+## Running group of tests with same label
```
lbarcziova, line 204:
```suggestion
## Running a group of tests with the same label
```

lbarcziova (approved):
thanks!!

content/docs/testing-farm.md
psss: Thanks for the heads-up. Very nice! Looking forward to trying this out :) Is it supposed to be enabled on the production instance today?

Frawless: The feature was merged last week so I think it should be in prod already.

PR #689, opened 2023-07-12
Install prebuild hugo binary from GitHub

jpopelka:
It's not in EPEL.

### DIFF
--- a/files/install-hugo.yaml
+++ b/files/install-hugo.yaml
@@ -3,10 +3,10 @@
   hosts: all
   tasks:
     - name: Install hugo
-      ansible.builtin.dnf:
-        name:
-          - hugo
-        state: latest
+      ansible.builtin.unarchive:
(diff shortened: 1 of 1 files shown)


mfocko (approved):
I should finish the Docusaurus migration, we wouldn't need Zuul :grin:

files/install-hugo.yaml
```diff
@@ -3,11 +3,10 @@
...
     - name: Install hugo
-      ansible.builtin.dnf:
-        name:
-          - hugo
-        state: latest
-      become: true
+      ansible.builtin.unarchive:
+        src: https://github.com/gohugoio/hugo/releases/download/v0.98.0/hugo_extended_0.98.0_Linux-64bit.tar.gz
+        remote_src: true
+        dest: /usr/bin/
```
mfocko, line 9: Btw, don't you need `become: true`, here? Since it's in `/usr/bin/`?

jpopelka: yup, that was it, thanks!

PR #693, opened 2023-07-14
feat(examples): add example of archive with submodules

mfocko:
Closes packit/packit#1992

### DIFF
--- a/content/docs/Configuration/examples.md
+++ b/content/docs/Configuration/examples.md
@@ -73,6 +73,16 @@ create-archive:
 ```
 {{< /details >}}
 
+{{< details title="Creating archive with git submodules" >}}
+```yaml
+create-archive:
+  - git submodule update --init
+  # Create an archive file with the ‹pkg› directory prefix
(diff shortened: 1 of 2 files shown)


lachmanfrantisek (approved):
Thanks!

content/docs/Configuration/examples.md
```diff
@@ -73,6 +73,16 @@ create-archive:
 ```
 {{< /details >}}
 
+{{< details title="Creating archive with git submodules" >}}
+```yaml
+create-archive:
+  - git submodule update --init
+  # Create an archive file with the ‹pkg› directory prefix
+  - bash -c 'git ls-files --recurse-submodules | tar --transform "s|^|pkg/|" -caf .packit/pkg.tar.gz -T-'
+  - echo '.packit/pkg.tar.gz'
```
junaruga, line 82: I think you need to remove `.packit/` in the 2 commands above too. In my case, there is `.packit` directory` in the repository. But there is not `.packit` directory generally in a repository.

PR #696, opened 2023-07-17
ci: fix token permissions for deploying

mfocko:


mfocko:
I guess it's late now :/ don't sue me :eyes:

PR #702, opened 2023-07-20
fix: enable usage charts

mfocko:


.github/workflows/deploy.yml
```diff
@@ -21,6 +21,11 @@ jobs:
       - name: Import the nested docs
         run: make import
 
+      - name: Generate newer usage graphs
+        run: |
+          pip3 install click requests pygal
+          python3 files/generate_usage.py
```
lachmanfrantisek, line 27: Nice! I don't need to update manually..;)

lbarcziova (approved):
:clap:

lachmanfrantisek (approved):

.github/workflows/deploy.yml
mfocko: It will be a bit messy when running locally. It seems to run with each `make run-dev`, maybe we'll need to adjust it a bit.

mfocko:
GH Action has been changed, need to merge manually…

PR #705, opened 2023-07-24
pull-from-upstream: document retrigger in job definition

TomasTomecek:
this was the first place I looked for it


mfocko (approved):
We should probably support api tokens and triggering via CLI

PR #714, opened 2023-08-15
Fix example of changelog-entry action

lbarcziova:
As discused in https://github.com/packit/packit/issues/2042, this is needed so that the macro is expanded.


nforro (approved):
Thanks!

lachmanfrantisek (approved):

PR #735, opened 2023-09-12
Document notifications.failure_comment.message

lbarcziova:
Related to packit/packit-service#1911

Merge once in production.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -387,21 +387,50 @@ repository.  Defaults to 4, that is, patches will look like `PatchNNNN:
 minimum width".
 
 #### notifications
+##### pull_request
+###### successful_build
+(*bool*)
+Enable comment with instructions how to install a package with the change implemented in the pull request. 
+This comment will be posted by Packit after a successful build of a pull request is done. 
 
-There is only one notification configuration you can set up right now: enable
-the "Congratulations!" comment which will packit send after a successful build of a
-pull request is done.
-
-The default behaviour is not to send the comment with instructions how to install a
-package with the change implemented in the pull request:
+The default behaviour is not to send the comment. To enable the comment, this should be the configuration:
 
 ```yaml
 notifications:
   pull_request:
-    successful_build: false
+    successful_build: true
 ```
 
-You can enable the commenting by setting `successful_build` to `true`.
+##### failure_comment
+###### message
+(*string*)
+A message that will be posted as a comment in case of a job failure. 
+Same as for other keys, you can define this option at the top-level, applying it to all jobs, or configure it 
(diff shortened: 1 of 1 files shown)


docs/configuration/index.md
```diff
@@ -374,21 +374,48 @@ repository.  Defaults to 4, that is, patches will look like `PatchNNNN:
...
-There is only one notification configuration you can set up right now: enable
-the "Congratulations!" comment which will packit send after a successful build of a
-pull request is done.
-
-The default behaviour is not to send the comment with instructions how to install a
-package with the change implemented in the pull request:
+##### pull_request
+###### successful_build
+(*bool*)
+Enable "Congratulations!" comment which will be posted by Packit after a successful build of a
```
lachmanfrantisek, line 380: I'm not sure if people still understand the "Congratulations!" since most of the current users can't remember this being the default behaviour. But since it's quite well described later on, we can leave it there.

docs/configuration/index.md
```diff
@@ -374,21 +374,48 @@ repository.  Defaults to 4, that is, patches will look like `PatchNNNN:
...
 ```
 
-You can enable the commenting by setting `successful_build` to `true`.
+##### failure_comment
+###### message
+(*string*)
+A message that will be posted as a comment in case of a job failure. 
+Same as for other keys, you can define this option at the top-level, applying it to all jobs, or configure it 
+on a job level, enabling you to have specific messages for each job or opt not to have a message for certain 
+jobs. By default, no message is posted on job failure.
```
lachmanfrantisek, line 396: Good point with the top-level/job-level reminder. People still don't know this.

docs/configuration/index.md
```diff
@@ -374,21 +374,48 @@ repository.  Defaults to 4, that is, patches will look like `PatchNNNN:
...
+    targets:
+      - fedora-all
+
+  - job: tests
+    trigger: pull_request
+    targets:
+      - fedora-all
+    notifications:
+      failure_comment:
+        message: "One of the tests failed for {commit_sha}, @admin, check the test failures on this PR."
```
lachmanfrantisek, line 417: +1 for using the user/group tag in the example. Do you think we should also mention this idea/usecase in the text?

lbarcziova: good idea, added

martinpitt:
The feature landed and was announced -- time to land this documentation as well? Thanks!

lbarcziova:
@martinpitt oh sure, thanks for reminding!

lachmanfrantisek (approved):

lbarcziova:
regate

martinpitt:
I see it in https://packit.dev/docs/configuration/#message now, thanks!

PR #737, opened 2023-09-13
fix(actions): use correct name of the variables

mfocko:
Noticed it when setting up `pull-from-upstream` on my package. Please don't ask how did I notice it…

You can see them being passed here: https://github.com/packit/packit/blob/80d9002ef29c69c5190b21b33b8624a77dc5c897/packit/config/common_package_config.py#L475-L483


lachmanfrantisek (approved):
Well spotted!

PR #740, opened 2023-09-14
pull-from-upstream: escape '-' for echo

TomasTomecek:
otherwise echo would interpret it


nforro (approved):

TomasTomecek:
Works as expected: https://src.fedoraproject.org/rpms/alot/pull-request/3#request_diff

PR #743, opened 2023-09-18
"Execute something in dist-git" example

majamassarini:
Merge after packit/packit#2081 and packit/sandcastle#197

### DIFF
--- a/docs/configuration/actions.md
+++ b/docs/configuration/actions.md
@@ -268,3 +268,24 @@ actions:
 
 Please note that `%changelog` expects dashes in a changelog entry, but the dash
 would be interpreted by `echo` as an option so we need to quote it.
+
+-----
+Be aware, that commands do not share the working directory.
+So, if you want to change the working directory of a command or multiple commands,
+don't run `cd` as a separate command as shown in the following example.
+This won't work as expected:
+
+```yaml
+actions:
+  post-upstream-clone:
+    - cd /tmp
+    - touch test_file
+```
+
+Instead, spawn a shell and change the directory as the first command:
+
+```yaml
+actions:
+  post-upstream-clone:
+    - bash -c "cd ${PACKIT_DOWNSTREAM_REPO}; touch new_file_for_dist_git.tmp"
+```


docs/configuration/actions.md
```diff
@@ -268,3 +268,13 @@ actions:
 
 Please note that `%changelog` expects dashes in a changelog entry, but the dash
 would be interpreted by `echo` as an option so we need to quote it.
+
+-----
+If you want to execute something in a different directory then you need to specify 
+all the commands inside the same directive:
```
nforro, line 274:
```suggestion
Commands default to upstream git repo as a working directory. If you want to change it (to a dist-git repo for example), spawn a shell and change the directory as the first command:
```

majamassarini: I also wanted to mention that upstream is the default execution dir but this is not always true. So I tried to be more generic, since this paragraph is not strictly related with propose_downstream and pull_from_upstream.

nforro: Ok, then something like this:

```suggestion
If you want to change the working directory of a command or multiple commands, don't do this, it won't work as expected:

```yaml
actions:
  post-upstream-clone:
    - cd /tmp
    - touch test_file"
\```

Instead, spawn a shell and change the directory as the first command:
```

docs/configuration/actions.md
```diff
@@ -268,3 +268,21 @@ actions:
...
+
+-----
+If you want to change the working directory of a command or multiple commands, don't do this, it won't work as expected:
+
+```yaml
+actions:
+  post-upstream-clone:
+    - cd /tmp
+    - touch test_file"
+\```
```
nforro, line 280: Sorry, I had to add the backslash there for the suggestion not to end too early.

docs/configuration/actions.md
```diff
@@ -268,3 +268,21 @@ actions:
...
+
+Instead, spawn a shell and change the directory as the first command:
+
+```yaml
+actions:
+  post-upstream-clone:
+    - bash -c "cd ${PACKIT_DOWNSTREAM_REPO}; touch new_file_for_dist_git.tmp"
```
nforro, line 287: I'm thinking if we can make both examples equivalent. It's not really possible to use `cd ${PACKIT_DOWNSTREAM_REPO}` in the first one, because the variable is expanded only in a shell and wrapping `cd` in `bash` is obvious nonsense. Do you think it's fine like this?

majamassarini: I am not following you 😅

nforro: This is nonsense, because `cd` in a subshell only affects that subshell:

```yaml
actions:
  post-upstream-clone:
    - bash -c "cd ${PACKIT_DOWNSTREAM_REPO}"
    - touch new_file_for_dist_git.tmp
```

This is not a good example, because `PACKIT_DOWNSTREAM_REPO` wouldn't be expanded:

```yaml
actions:
  post-upstream-clone:
    - cd ${PACKIT_DOWNSTREAM_REPO}
    - touch new_file_for_dist_git.tmp
```

I just thought it would be nicer to have both examples do the exact same thing, just the first one in a way that doesn't work in practice. But I guess it's fine like it is :slightly_smiling_face:

nforro (approved):

docs/configuration/actions.md
nforro: Yes, but ideally this wouldn't be needed and people could just write:

```yaml
actions:
  post-upstream-clone:
    script: |
      cd ${PACKIT_DOWNSTREAM_REPO}
      touch new_file_for_dist_git.tmp
```

We have an open issue for that: https://github.com/packit/packit/issues/1641

docs/configuration/actions.md
```diff
@@ -268,3 +268,21 @@ actions:
 
 Please note that `%changelog` expects dashes in a changelog entry, but the dash
 would be interpreted by `echo` as an option so we need to quote it.
+
+-----
+If you want to change the working directory of a command or multiple commands, don't do this, it won't work as expected:
```
lachmanfrantisek, line 273:
```suggestion
Be aware, that commands do not share the working directory.
So, if you want to change the working directory of a command or multiple commands,
don't run `cd` as a separate command as shown in the following example.
This won't work as expected:
```

lachmanfrantisek (approved):

PR #745, opened 2023-09-18
Add blogpost for week 37

majamassarini:
Should we mention that we had the workshop and link the video?


lachmanfrantisek (approved):
Thanks for doing the move two times!

PR #753, opened 2023-10-03
Clarify manual_trigger enabled for both build and tests

lbarcziova:
Related to packit/packit-service#2197

### DIFF
--- a/docs/configuration/upstream/tests.md
+++ b/docs/configuration/upstream/tests.md
@@ -70,6 +70,13 @@ If you want to trigger tests via pull request comment and not by every new commi
     manual_trigger: true
 ```
 
+:::info
+
+If you have both build and follow-up test jobs with `manual_trigger: true`, you will need to post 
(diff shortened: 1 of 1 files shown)


docs/configuration/upstream/tests.md
```diff
@@ -70,6 +70,13 @@ If you want to trigger tests via pull request comment and not by every new commi
     manual_trigger: true
 ```
 
+:::info
+
+If you have both build and followup test jobs with `manual_trigger: true`, you will need to post both `/packit build` and
+after the builds are successful `/packit test` comment commands.
```
xsuchy, line 76: I had to read it several times to understand it. I would suggest:
"...you will need to post both: first post `/packit build` command and
after the builds are successful post  `/packit test` comment command."

And nit-pick - s/followup/follow-up/ but some dictionaries recognize "followup" as a word.

lbarcziova: good points, updated

xsuchy (approved):
+1

PR #754, opened 2023-10-04
pull-from-uptream - add link to Dashboard

xsuchy:
Otherwise user has no clue what is going next. And even worse, if the job fails, he do not know how to debug it.


lbarcziova (approved):
thanks! 

To me, it would make sense to include it (also) somewhere in the first section since it doesn't apply only to retriggering, I will leave that up to you

PR #769, opened 2023-11-03
Add more customizable details for comment failures

majamassarini:
Merge after packit/packit-service#2245


docs/configuration/index.md
```diff
@@ -411,8 +411,13 @@ jobs. By default, no message is posted on job failure.
...
 To prevent duplication,  Packit posts a comment only when its content differs from the previous comment in the specific 
 pull request or commit.
-To include dynamic content, you can use the `{commit_sha}` placeholder, which will be automatically replaced with 
-the actual commit SHA provided by Packit (consider using this in relation to the duplication of comments). 
+To include dynamic content, you can use multiple placeholders which will be automatically replaced (consider using this in relation to the duplication of comments):
+
+  - `{commit_sha}`: replaced with the actual commit SHA provided by Packit
+  - `{logs_url}`: replaced with the url to the service's logs, the service can be Copr, Koji or Testing Farm depending on the Packit job
+  - `{packit_dashboard_url}`: replaced with the Packit dashboard url for the job
+  - `{external_dashboard_url}`: url to the service dashboard, the service can be Copr, Koji or Testing Farm depending on the Packit job
```
lachmanfrantisek, line 419: Do you think we should provide some real example URLs?

lachmanfrantisek (approved):

lbarcziova (approved):

docs/configuration/index.md
majamassarini: I don't know. If I point to the copr logs as an example I fear they will disappear after a while or not?

lbarcziova: even if the links become invalid in the future, I think it could at least give users the idea of what kind of URL we are talking about, so I would be still for adding some examples

PR #777, opened 2023-11-20
Document pre-releases

nforro:


docs/pre-releases.md
```diff
@@ -0,0 +1,82 @@
...
+
+Source0:        https://example.com/files/v%{majorver}.%{minorver}/%{name}-%{upstream_version}.tar.gz
+
+...
+
+%prep
+%autosetup -p1 -n %{name}-%{upstream_version}
+```
+
+Packit supports both ways. There is a config option `prerelease_suffix_pattern` that defines a regex that is used to determine if a version is a pre-release or not. It defaults to `([.\-_~^]?)(alpha|beta|rc|pre(view)?)([.\-_]?\d+)?` and you can adjust it if your project uses a different versioning scheme or undefine it to opt-out from pre-release processing. The first capturing group of the regex should capture the delimiter and can be empty if there is none.
```
lbarcziova, line 81: could you please also add the options to the configuration docs and link it from here?

lbarcziova (commented):
otherwise LGTM :+1:

docs/pre-releases.md
nforro: Done, I hope I didn't mess up the links.

lbarcziova (approved):
nicelly done!

PR #779, opened 2023-11-22
Clarify issue_repository config option

### LINKED ISSUE #771: packit.yaml/issue_repository is confusing for pull-from-upstream
### What happened? What is the problem?

I got this report 
  https://github.com/xsuchy/python-hwdata/issues/5
that pull-from-upstream failed. Put aside why it failed. My problem is that pull-from-upstream failures should not be reported to upstream. In this case, I am the owner of upstream as well. But if upstream would be somebody else, they can be vary angry about it ("why you are bothering us about it, not our business").

This happened because I put `issue_repository` in packit.yaml in dist-git:
https://src.fedoraproject.org/rpms/python-hwdata/blob/rawhide/f/packit.yaml#_3
so one can argue it is my fault and PEBKAC. 

But it was nowhere mentioned that dist-git import failures will be reported to issue_repository. And I checked
 https://packit.dev/docs/configuration/downstream/pull_from_upstream#first-setup
https://packit.dev/docs/configuration#issue_repository

I propose:

[ ] amend https://packit.dev/docs/configuration#issue_repository alter last two sentences of the first paragraph to use for issue_repository the downstream repository. Maybe even with an exclamation mark.
[ ] alter https://packit.dev/docs/configuration/downstream/pull_from_upstream#first-setup to include in Caution box that you SHOULD NOT use the upstream issue repository there.
[ ] amend https://packit.dev/docs/configuration/downstream/pull_from_upstream#example to include downstream issue repository

### What did you expect to happen?

_No response_

### Example URL(s)

_No response_

### Steps to reproduce

```bash
1.
2.
3.
```

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

lbarcziova:
Fixes #771

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -448,14 +448,10 @@ this message is used as an extension of the comment added by default by Packit. 
 #### issue_repository
 
 Use this key to be notified about errors of the downstream jobs (Koji build, Bodhi update, pull from upstream).
-The value can be a URL of any GitHub/GitLab/Pagure project including dist-git
-where issues are enabled and Packit has an identity on that git forger instance.
+The value can be a URL of any GitHub/GitLab/Pagure project where issues are enabled and Packit has an identity on that git forge instance.
 Alternatively, you can set up a dedicated project for receiving such notifications.
 (Let us know if you need another instance to be supported.)
-It does not need to be an upstream repository,
-you can also enable an issue tracker for the dist-git repository
-and use this as a place for those notifications.
-Or, you can set up a dedicated project for receiving such notifications.
+It does not need to be an upstream repository (and in case you do not maintain it, it also **should not** be).
 
 By default, no issue will be created so Packit doesn't unintentionally spam any repository.
 (E.g. Since those jobs are defined in downstream, upstream project does not need to be aware of Packit.)
--- a/docs/configuration/downstream/bodhi_update.md
+++ b/docs/configuration/downstream/bodhi_update.md
@@ -43,8 +43,9 @@ or any improvement we can implement.
   to not need to change the config file when the new system version is released.
 
 ## Example
-
 ```yaml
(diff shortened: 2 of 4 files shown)


docs/configuration/index.md
```diff
@@ -449,11 +449,11 @@ this message is used as an extension of the comment added by default by Packit.
...
 Use this key to be notified about errors of the downstream jobs (Koji build, Bodhi update, pull from upstream).
 The value can be a URL of any GitHub/GitLab/Pagure project including dist-git
-where issues are enabled and Packit has an identity on that git forger instance.
+where issues are enabled and Packit has an identity on that git forge instance.
 Alternatively, you can set up a dedicated project for receiving such notifications.
 (Let us know if you need another instance to be supported.)
-It does not need to be an upstream repository,
-you can also enable an issue tracker for the dist-git repository
+It does not need to be an upstream repository (and in case you do not maintain it, it also **should not** be).
+You can also enable an issue tracker for the dist-git repository
```
nforro, line 456: Didn't we find out that this is in fact not possible? :sweat_smile:

lbarcziova: I was not 100% sure if I remembered that correctly :D whether it wasn't impossible only via UI..I will remove it then

docs/configuration/index.md
```diff
@@ -449,13 +449,10 @@ this message is used as an extension of the comment added by default by Packit.
 
 Use this key to be notified about errors of the downstream jobs (Koji build, Bodhi update, pull from upstream).
 The value can be a URL of any GitHub/GitLab/Pagure project including dist-git
```
nforro, line 451: Sorry for nitpicking, but we should probably remove `including dist-git` from here as well.

lbarcziova:
regate

PR #789, opened 2023-12-07
Document new test_command options

lbarcziova:
Documentation for packit/packit-service#2270
Related to packit/packit#2056

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -563,6 +563,39 @@ The value requires a Python regex pattern and is used with `re.match` function.
  matched values are not the same, thus Packit will not create a *3.9.0*  pull request release on dist-git target 
  branch. If there is an *3.8.1* release, the matched values (=`3.8`) are the same and pull request is created.
 
+
+#### test_command
+##### default_identifier
+The identifier (refer to [`identifier`](./configuration/upstream/tests#optional-parameters)) utilised by default when
+the `/packit test` comment command is invoked without arguments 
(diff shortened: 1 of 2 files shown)


docs/configuration/index.md
```diff
@@ -563,6 +563,39 @@ The value requires a Python regex pattern and is used with `re.match` function.
...
+```yaml
+default_labels:
+  - regression
+  - upgrade
+```
+As a result, invoking `/packit test` will be automatically interpreted as `/packit test --labels regression,upgrade`.
+
+:::info
+`default_identifier` and `default_labels` are not supposed to be used simultaneously. You should configure
+only on of these options.
```
nforro, line 597:
```suggestion
only one of these options.
```

docs/configuration/index.md
```diff
@@ -563,6 +563,39 @@ The value requires a Python regex pattern and is used with `re.match` function.
...
+you can configure the following:
+```yaml
+default_labels:
+  - regression
+  - upgrade
+```
+As a result, invoking `/packit test` will be automatically interpreted as `/packit test --labels regression,upgrade`.
+
+:::info
+`default_identifier` and `default_labels` are not supposed to be used simultaneously. You should configure
```
nforro, line 596: Shouldn't we be more strict here? What happens if a user configures both?

```suggestion
`default_identifier` and `default_labels` can't be used simultaneously. You can configure
```

lbarcziova: well, they can be configured simultaneously, nothing will break, but only one of them will be taken into consideration

nforro (approved):

PR #792, opened 2024-01-02
chore(weekly): add last blog of ‘23

mfocko:


weekly/2023/week-52.md
```diff
@@ -0,0 +1,16 @@
...
+date: 2024-01-02
+tags:
+  - 2023-December
+  - 2023
+  - December
+---
+
+## End of the year in Packit
+
+- We have hit some issues with the firewall rules on the new cluster hosting our production. The fixes will be deployed around January 11th. At the moment we are aware of the issues related to the:
```
lbarcziova, line 13:
```suggestion
- We have hit some issues with the firewall rules on the new cluster hosting our production. We are waiting for the required fix that is done outside of the Packit team which should be deployed around January 11th. At the moment we are aware of the issues related to the:
```

or something similar to express that we cannot do much about it

lbarcziova: and we could also add something like: `We will keep you informed and let you know once the fix is deployed.`

weekly/2023/week-52.md
```diff
@@ -0,0 +1,16 @@
...
+date: 2024-01-02
+tags:
+  - 2023-December
+  - 2023
+  - December
+---
+
+## End of the year in Packit
+
+- We have hit some issues with the firewall rules on the new cluster hosting our production. We are waiting for the required fix that is done outside of our team which should be deployed around January 11th. We will keep you informed about the curretn status. At the moment we are aware of the issues related to the:
```
lbarcziova, line 13:
```suggestion
- We have hit some issues with the firewall rules on the new cluster hosting our production. We are waiting for the required fix that is done outside of our team which should be deployed around January 11th. We will keep you informed about the current status. At the moment we are aware of the issues related to the:
```

lbarcziova (approved):
thanks!

PR #793, opened 2024-01-03
Fix authors for blog week 49,50

lbarcziova:


lbarcziova:
I know, but for week 50 the typo caused it is now rendered like this and it hurts my eyes (and as I touched it, I also added it for week 49 for consistency)

Snímka obrazovky z 2024-01-03 14-23-20

PR #799, opened 2024-01-09
State explicitly where we take the tarball URL

majamassarini:


docs/guide.md
```diff
@@ -352,6 +352,9 @@ When you are not allowed or do not want to run packit command from the upstream
 git repository, you can run commands from the cloned dist-git repository in the
 same way as you do from upstream.  Make sure, that you specify the
 `upstream_project_url` in your configuration.
+You don't need to specify the upstream project tarball URL but just the upstream
+ project repo URL in the Packit configuration file because the tarball URL is taken 
+ from the specfile when needed.
```
lbarcziova, line 357: I am not sure how relevant is it to the question here :thinking: but I would definitely mention it in the Fedora release guide

docs/configuration/index.md
```diff
@@ -115,7 +115,8 @@ Previously known as `upstream_project_name`, which has been deprecated since Pac
 #### upstream_project_url
 
 (*string*) URL of the upstream git project (e.g.
-`https://github.com/packit/packit`).
+`https://github.com/packit/packit`). This isn't the URL to the upstream project tarball.
+The URL to the upstream project tarball is taken from the specfile when needed.
```
nforro, line 119: This is not really true for upstream jobs. There the tarball is generated and a potential URL in spec file is overwritten with it.

```suggestion
`https://github.com/packit/packit`).

:::info

Do not confuse this with the URL of upstream tarball. In case of `propose_downstream`, `pull_from_upstream`, `koji_build` and `bodhi_update` jobs, upstream tarball URL is taken from spec file or from [sources](#sources).

:::
```

docs/guide.md
nforro, line 357: I don't think this should be here. This paragraph seems to talk about SRPM building.

majamassarini:
recheck

lbarcziova (approved):
thanks!

nforro (approved):

PR #814, opened 2024-01-25
Document support for groups and aliases in koji allowed_* options

lbarcziova:
Related to packit/packit#2088

### DIFF
--- a/docs/configuration/downstream/koji_build.md
+++ b/docs/configuration/downstream/koji_build.md
@@ -47,11 +47,20 @@ For Koji builds from upstream, see [`upstream_koji_build`](/docs/configuration/u
 
 * **scratch** - defaults to `false`, use to create scratch (test) builds
   instead of the real production builds
-* **allowed_pr_authors** - a list of FAS accounts of PR authors whose merged pull requests will trigger koji builds
-  (defaults to `['packit']`).
-* **allowed_committers** - a list of FAS accounts of committers whose direct pushes to dist-git will trigger koji builds
-  (defaults to an empty list). You do not need to configure this option if you want to have koji builds triggered only by merged pull requests.
-
+* **allowed_pr_authors** - a list of PR authors whose merged pull requests will trigger Koji builds
+  (defaults to `['packit']`). As each item of the list you can specify:
+   - name of a FAS account, e.g. `my-fas`
+   - name of a group - starting with `@`, e.g. `@my-sig`
+   - `all_admins` alias - allowing all users with admin access to the dist-git repo
+   - `all_committers` alias - allowing all users with commit access to the dist-git repo
+* **allowed_committers** - a list of committers whose direct pushes to dist-git will trigger Koji builds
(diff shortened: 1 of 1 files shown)


docs/configuration/downstream/koji_build.md
```diff
@@ -47,11 +47,18 @@ For Koji builds from upstream, see [`upstream_koji_build`](/docs/configuration/u
...
-* **allowed_pr_authors** - a list of FAS accounts of PR authors whose merged pull requests will trigger koji builds
-  (defaults to `['packit']`).
-* **allowed_committers** - a list of FAS accounts of committers whose direct pushes to dist-git will trigger koji builds
-  (defaults to an empty list). You do not need to configure this option if you want to have koji builds triggered only by merged pull requests.
-
+* **allowed_pr_authors** - a list of PR authors whose merged pull requests will trigger Koji builds
+  (defaults to `['packit']`). You can specify:
+   - FAS accounts directly, e.g. `my-fas`
+   - groups - starting with `@`, e.g. `@my-sig`
+   - aliases `all_admins` and `all_committers` - allowing all users with corresponding access to the dist-git repo
```
majamassarini, line 54: I am just wondering if someone will think how to substitute `all_admin` and `all_committers`.
Since the previous two examples were meant to be substituted, like `my-fas` with `mmassari`. I don't want that someone thinks `all_admins` should be substituted with a list of admins... Just to be sure, I will put directly a user example and a group example in the fas and group lines.

majamassarini (approved):

docs/configuration/downstream/koji_build.md
nforro: I think it's pretty clear, but maybe something like this would be a bit better?

```suggestion
   - name of a FAS account, e.g. `my-fas`
   - name of a group - starting with `@`, e.g. `@my-sig`
   - `all_admins` alias - allowing all users with admin access to the dist-git repo
   - `all_committers` alias - allowing all users with commit access to the dist-git repo
```

lbarcziova: thanks for the feedback, the suggestion makes sense to me

docs/configuration/downstream/koji_build.md
```diff
@@ -47,11 +47,20 @@ For Koji builds from upstream, see [`upstream_koji_build`](/docs/configuration/u
 
 * **scratch** - defaults to `false`, use to create scratch (test) builds
   instead of the real production builds
-* **allowed_pr_authors** - a list of FAS accounts of PR authors whose merged pull requests will trigger koji builds
-  (defaults to `['packit']`).
-* **allowed_committers** - a list of FAS accounts of committers whose direct pushes to dist-git will trigger koji builds
-  (defaults to an empty list). You do not need to configure this option if you want to have koji builds triggered only by merged pull requests.
-
+* **allowed_pr_authors** - a list of PR authors whose merged pull requests will trigger Koji builds
+  (defaults to `['packit']`). You can specify:
```
nforro, line 51: Ok, now I'd suggest one more change:

```suggestion
  (defaults to `['packit']`). As each item of the list you can specify:
```

PR #815, opened 2024-01-26
Document propose-downstream for CentOS Stream

### LINKED ISSUE #796: Document the ‹propose-downstream› workflow for CentOS Stream
Document the way to enable `propose-downstream` workflow for the CentOS Stream. This must be done using 2 different packages (since they don't share packaging tool, and it's also beneficial to have them separate because of potentially different specfiles being used).

Example of such config:
```yaml
---
downstream_package_name: nispor
upstream_project_url: https://github.com/nispor/nispor
issue_repository: https://github.com/nispor/nispor
update_release: false
upstream_package_name: nispor
upstream_tag_template: v{version}
copy_upstream_release_description: false

packages:
  nispor-fedora:
    actions:
      post-upstream-clone:
        - wget https://src.fedoraproject.org/rpms/nispor/raw/rawhide/f/nispor.spec
  nispor-centos:
    pkg_tool: centpkg
    actions:
      post-upstream-clone:
        - wget https://gitlab.com/redhat/centos-stream/rpms/nispor/-/raw/c9s/nispor.spec

jobs:
  - job: propose_downstream
    trigger: release
    packages: [nispor-fedora]
    dist_git_branches:
      - fedora

  - job: propose_downstream
    trigger: release
    packages: [nispor-centos]
    dist_git_branches:
      - c9s
```

lbarcziova:
Fixes #796

### DIFF
--- a/docs/configuration/upstream/propose_downstream.md
+++ b/docs/configuration/upstream/propose_downstream.md
@@ -5,8 +5,15 @@ sidebar_position: 5
 
 # `job: propose_downstream`
 
-Land a new upstream release in Fedora. This job only makes sure the changes
-happen in Fedora dist-git - no builds. A pull request is created as a result.
+Land a new upstream release in Fedora (or CentOS Stream). This job only makes sure the changes
+happen in dist-git - no builds. A pull request is created as a result.
+
+:::tip CentOS Stream support
+
+We newly support `propose-downstream` also for CentOS Stream. The workflow is the same as for Fedora, but there are few
+configuration specifics, see [the subsection below](#syncing-the-release-to-centos-stream).
+
+:::
 
 ## Supported triggers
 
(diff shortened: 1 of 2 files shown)


docs/configuration/upstream/propose_downstream.md
```diff
@@ -32,7 +40,87 @@ jobs:
...
+pkg_tool: centpkg
+actions:
+  post-upstream-clone:
+    - wget https://gitlab.com/redhat/centos-stream/rpms/nispor/-/raw/c9s/nispor.spec
+
+jobs:
+  - job: propose_downstream
+    trigger: release
+    dist_git_branches:
+      - c9s
```
lbarcziova, line 81: or do you think this is confusing and we should just directly advise having the 2 packages below?

docs/configuration/upstream/propose_downstream.md
```diff
@@ -17,6 +23,8 @@ happen in Fedora dist-git - no builds. A pull request is created as a result.
 * **dist_git_branches** - a (list of) branches in dist-git where packit should work (defaults to `main` which represents _Fedora Rawhide_).
   You can also use the [aliases provided by Packit](/docs/configuration#aliases)
   to not need to change the config file when the new system version is released.
+* **pkg_tool** - `fedpkg` (default) or `centpkg`. Needs to be specified if you want to sync release to CentOS Stream,
```
mfocko, line 26: Unless you've opened PR for this, we don't support it. The issue is that the `LocalProject` is initialized with the `pkg_tool`, so you need to have it specified **before** the job. We probably “hack” it somehow, but that introduces another problem of `LocalProject` depending on the job config during initialization.

> **NOTE**
> I _THINK_ it doesn't work :D

docs/configuration/upstream/propose_downstream.md
mfocko: I don't really mind it, the problem is that we cannot _overload_ the config in the Service, so this looks nice, but you can only use it locally and with explicitly specified config (in case you have Fedora set up too) :/

Maybe with a disclaimer that this can be used as a minimal config with CLI with `--config` switch

docs/configuration/upstream/propose_downstream.md
```diff
@@ -3,10 +3,16 @@ title: Propose downstream
...
-Land a new upstream release in Fedora. This job only makes sure the changes
-happen in Fedora dist-git - no builds. A pull request is created as a result.
+Land a new upstream release in Fedora (or CentOS Stream). This job only makes sure the changes
+happen in dist-git - no builds. A pull request is created as a result.
+
+:::tip CentOS Stream support
+
+We newly support `propose-downstream` also for CentOS Stream, see the configuration specifics in the 
+[subsection below](#syncing-the-release-to-centos-stream).
+:::
```
mfocko, line 15:
```suggestion
:::tip CentOS Stream support

We newly support `propose-downstream` also for CentOS Stream, see the configuration specifics in the 
[subsection below](#syncing-the-release-to-centos-stream).

:::
```

Empty lines are because of `prettier` potentially turning:
```
:::tip
something
:::
```
into
```
:::tip something :::
```
which breaks the parser…

mfocko (commented):
LGTM

docs/configuration/upstream/propose_downstream.md
lbarcziova: aah ok, I wasn't aware of this, will change it

docs/configuration/upstream/propose_downstream.md
lbarcziova: hmm, in that case, I would rather remove it to not create unnecessary confusion

docs/configuration/upstream/propose_downstream.md
```diff
@@ -32,7 +39,63 @@ jobs:
...
-This config would update Fedora Rawhide and Fedora 35 dist-git branches.
+This config would update Fedora Rawhide and Fedora 39 dist-git branches.
+
+
+
+## Syncing the release to CentOS Stream
+
+Recently, we have implemented support for `propose-downstrem` also for CentOS Stream. This means that you can now configure
+the upstream release to trigger release syncing the same way as for Fedora to CentOS Stream dist-git repositories
+hosted in https://gitlab.com/redhat/centos-stream/rpms.
```
lachmanfrantisek, line 53: Do we want to provide a disclaimer/warning that it might not work as expected?

docs/configuration/upstream/propose_downstream.md
```diff
@@ -5,8 +5,15 @@ sidebar_position: 5
...
 
-Land a new upstream release in Fedora. This job only makes sure the changes
-happen in Fedora dist-git - no builds. A pull request is created as a result.
+Land a new upstream release in Fedora (or CentOS Stream). This job only makes sure the changes
+happen in dist-git - no builds. A pull request is created as a result.
+
+:::tip CentOS Stream support
+
+We newly support `propose-downstream` also for CentOS Stream, see the configuration specifics in the 
+[subsection below](#syncing-the-release-to-centos-stream).
```
lachmanfrantisek, line 14:
```suggestion
We newly support `propose-downstream` also for CentOS Stream. The workflow is the same as for Fedora, but see the configuration specifics in the 
[subsection below](#syncing-the-release-to-centos-stream).
```

Or something like that to let people know that the process of the sync itself is the same.

docs/configuration/upstream/propose_downstream.md
```diff
@@ -32,7 +39,63 @@ jobs:
 - job: propose_downstream
   trigger: release
   dist_git_branches:
-    - f35
+    - f39
 ```
```
lachmanfrantisek, line 43: :smile:

docs/configuration/upstream/propose_downstream.md
lachmanfrantisek: Also, what about asking for feedback and reporting issues so we can improve the CentOS integration?

lachmanfrantisek (approved):
Thanks a lot! The config does not look so scary..;)

docs/configuration/upstream/propose_downstream.md
lbarcziova: good point, I will add a hint or something like that

mfocko (approved):

lbarcziova:
regate

PR #820, opened 2024-01-29
Document parse_time_macros

nforro:
Related to https://github.com/packit/packit/pull/2222.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -628,6 +628,30 @@ As a result, invoking `/packit test` will be automatically interpreted as `/pack
 `default_identifier` and `default_labels` are not supposed to be used simultaneously. You should configure
 only one of these options.
 :::
+
+#### parse_time_macros
+
+(*dict*) Macros to be explicitly defined or undefined at spec file parse time.
+To undefine a macro, set its value to `null`.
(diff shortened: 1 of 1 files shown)


majamassarini (approved):

docs/configuration/index.md
```diff
@@ -628,6 +628,30 @@ As a result, invoking `/packit test` will be automatically interpreted as `/pack
 `default_identifier` and `default_labels` are not supposed to be used simultaneously. You should configure
 only one of these options.
 :::
+
+#### parse_time_macros
+
+(*dict*) Macros to be explicitly defined or undefined at spec file parse time.
+To undefine a macro, set its value to `null`.
+This can be useful for instance when the spec file is OS-specific and Packit runs on a different OS.
```
lbarcziova, line 636: > Packit runs on a different OS.

We could mention here what Packit runs on exactly.

lbarcziova (approved):

docs/configuration/index.md
nforro: You mean Packit Service, right? I was thinking about that, let me add it (but we should keep in mind to update it when we upgrade to c10s :slightly_smiling_face:).

PR #836, opened 2024-02-19
fix(docs/koji): add note about ‹scratch›

mfocko:
Reported by @pvalena during set up of upstream Koji builds for `dracut`.


lachmanfrantisek (approved):
Thanks!

PR #859, opened 2024-03-25
fix: yaml snippet for `test_command`

LecrisUT:
Quick documentation fix since the header only is not clear that `default_identifier` have to be under a `test_command` object


lbarcziova (approved):
lgtm, can you please rebase?

PR #861, opened 2024-03-28
Enhance description of `sync_changelog`

nforro:


LecrisUT (approved):

lbarcziova (approved):
thanks! But I am thinking about whether we shouldn't change this behaviour, it doesn't really align with the option name (we would need to check who is using this option).

LecrisUT:
`tmt` is using it. The thing is in the `%changelog` the verison/release numbers are used, so it makes sense to imply that `packit` does not update version, but I think packit should fail if it is not set to not update the version. Then intent is more specific.

martinhoyer:
> it doesn't really align with the option name 

+1. It is a bit misleading, even with the documentation change. 
Would it be hard to implement sync_changelog=true to not affect the other ways .spec is being processed, like Version, Release updates?
Although I'm not sure if it make sense - if a project is making .spec file changes outside of packit, it might as well handle the rest. Perhaps renaming it from sync_changelog to something that implies it's "syncing the whole spec" would be a way?

Why do projects, like tmt, don't use packit for generating changelog?

nforro:
> Perhaps renaming it from sync_changelog to something that implies it's "syncing the whole spec" would be a way?

Yes, this would make the most sense to me. Any ideas how to call it?

nforro:
Let's discuss this on the next architecture meeting. @martinhoyer in case you would like to participate, let me know and I'll invite you.

martinhoyer:
> Let's discuss this on the next architecture meeting. @martinhoyer in case you would like to participate, let me know and I'll invite you.

Thanks, I probably don't have much to add due to the lack of experience with packit (and in general), so all I had to say is in the above comment. To recap:
 - Is there a need for someone to only sync the changelog?
   - If yes, why? Would being able to customize the changelog generation in some way help?
   - If not, I'd just rename it.

PR #864, opened 2024-03-30
Use matrix.to in the link to Matrix room

### LINKED ISSUE #863: Matrix room link points to chat.fedoraproject Element
### What happened? What is the problem?

The matrix room mentioned in contact section - [#packit:fedora.im ](https://chat.fedoraproject.org/#/room/#packit:fedora.im)(Element / Matrix)
is directing users to fedoraproject.org web Element client explicitly.

### What did you expect to happen?

It is a common practice to use https://matrix.to, so users can easily join the room with their preferred client (and server, as the room is federated).  

### Example URL(s)

_No response_

### Steps to reproduce

_No response_

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [X] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

martinhoyer:
I don't really know how docusaurus (or the website in general) works and just replaced the links in both files I've found them. Hope it's ok.

Also changing "Element / Matrix" to just "Matrix", as matrix client is not relevant to the room.

Fixes
#863

RELEASE NOTES BEGIN

Matrix room link now uses matrix.to service

RELEASE NOTES END

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -275,7 +275,7 @@ const config = {
             items: [
               {
                 html: `
-                  <a href="https://matrix.to/#/%23packit:fedora.im">
+                  <a href="https://matrix.to/#/#packit:fedora.im?web-instance[element.io]=chat.fedoraproject.org">
                     #packit:fedora.im
                   </a> (Matrix)
                 `,
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -275,9 +275,9 @@ const config = {
             items: [
               {
                 html: `
-                  <a href="https://chat.fedoraproject.org/#/room/#packit:fedora.im">
+                  <a href="https://matrix.to/#/%23packit:fedora.im">
                     #packit:fedora.im
-                  </a> (Element / Matrix)
+                  </a> (Matrix)
                 `,
               },
               {
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -175,10 +175,10 @@ function Contacts() {
         <h2 id="contact">Contact</h2>
         <ul>
           <li>
-            <Link to="https://chat.fedoraproject.org/#/room/#packit:fedora.im">
+            <Link to="https://matrix.to/#/%23packit:fedora.im">
               #packit:fedora.im
             </Link>{" "}
-            (Element / Matrix)
+            (Matrix)
           </li>
           <li>
             <Link to="mailto:hello@packit.dev">hello@packit.dev</Link>
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -175,7 +175,7 @@ function Contacts() {
         <h2 id="contact">Contact</h2>
         <ul>
           <li>
-            <Link to="https://matrix.to/#/%23packit:fedora.im">
+            <Link to="https://matrix.to/#/#packit:fedora.im?web-instance[element.io]=chat.fedoraproject.org">
               #packit:fedora.im
             </Link>{" "}
             (Matrix)


src/pages/index.tsx
```diff
@@ -175,10 +175,10 @@ function Contacts() {
         <h2 id="contact">Contact</h2>
         <ul>
           <li>
-            <Link to="https://chat.fedoraproject.org/#/room/#packit:fedora.im">
+            <Link to="https://matrix.to/#/%23packit:fedora.im">
               #packit:fedora.im
             </Link>{" "}
-            (Element / Matrix)
+            (Matrix)
```
majamassarini, line 181:
```suggestion
            (Matrix - choose your app)
            <Link to="https://chat.fedoraproject.org/#/room/#packit:fedora.im">
              #packit:fedora.im
            </Link>{" "}
            (Matrix - Element app)            
```
We were wondering if having the old link can make life easier for users less used to matrix. So we were thinking of having them both.

majamassarini: Thanks for this useful summary. We will discuss it in the team once more and I will let you know.

lbarcziova: I think when adding this to our docs, we may have just copied this from how [Fedora docs](https://docs.fedoraproject.org/en-US/project/communications/#_primary_channels) have it. So I don't see any big issue with this. I would probably still prefer to provide both links so people can choose.

nforro: When you go to [https://matrix.to/#/#packit:fedora.im?web-instance[element.io]=chat.fedoraproject.org](https://matrix.to/#/#packit:fedora.im?web-instance%5Belement.io%5D=chat.fedoraproject.org), you can choose the _Element_ app (default choice) and there is _Open on chat.fedoraproject.org_ link which is the same as our "old" link. So IMHO it makes sense to have just one link.

martinhoyer: Exactly what Nikola said. Compared to the existing link, it's just one extra click to confirm client and server for users wishing to use Element web on Fedora server. :)

But honestly, please double check you want to use Fedora server as default there. I can explain how it all differs in more detail if you want.

docusaurus.config.js
```diff
@@ -275,9 +275,9 @@ const config = {
             items: [
               {
                 html: `
-                  <a href="https://chat.fedoraproject.org/#/room/#packit:fedora.im">
+                  <a href="https://matrix.to/#/%23packit:fedora.im">
```
majamassarini, line 278:
```suggestion
                  <a href="https://matrix.to/#/#packit:fedora.im?web-instance[element.io]=chat.fedoraproject.org">
```

src/pages/index.tsx
```diff
@@ -175,10 +175,10 @@ function Contacts() {
         <h2 id="contact">Contact</h2>
         <ul>
           <li>
-            <Link to="https://chat.fedoraproject.org/#/room/#packit:fedora.im">
+            <Link to="https://matrix.to/#/%23packit:fedora.im">
```
majamassarini, line 178:
```suggestion
            <Link to="https://matrix.to/#/#packit:fedora.im?web-instance[element.io]=chat.fedoraproject.org">
```

majamassarini (commented):
@martinhoyer we asked if we can redirect users to the fedora server (you correctly pointed out they are paying for those users we redirect). The answer they gave us follows:

> there is no policy about who can or cannot use it, and Packit is surely entitled to being under the Fedora umbrella if that is what the project wants. There is a lot of overlap already.
It is true that Fedora pays each user on the :fedora.im homeserver. But rooms/channels using :fedora.im or :fedoraproject.org have no cost.
We are having some issues with the Fedora homeserver for new account registration, so it might be wiser to use a :matrix.org account for now.
Plus, using :matrix.org means you can use the Element X mobile client, which is transformationally different and better than the normal Element mobile client 

For all this reasons and the above discussion we decided to use the matrix link which links to the fedoraserver as you can see in the review.

Thanks for bringing this up.

martinhoyer:
Thanks for looking into it.

> It is true that Fedora pays each user on the :fedora.im homeserver. But rooms/channels using :fedora.im or :fedoraproject.org have no cost.
We are having some issues with the Fedora homeserver for new account registration, so it might be wiser to use a :matrix.org account for now.
Plus, using :matrix.org means you can use the Element X mobile client, which is transformationally different and better than the normal Element mobile client

So you've decided against the Fedora suggestions? ;)

majamassarini (approved):
Why do you say it is against Fedora suggestions?

PR #867, opened 2024-04-05
Adding an example for extending a timeout for pipeline

see-quick:
This PR adds an example of Testing the farm timeout of the pipeline. The default value for pipeline is 12 hours but when one wants to extend such an attribute it needs to be configured in `tf_extra_params`. I have added this example because it was missing in packet DOCs.

### DIFF
--- a/docs/configuration/examples.md
+++ b/docs/configuration/examples.md
@@ -432,6 +432,30 @@ jobs:
 </details>
 
 <details>
+  <summary>Extending timeout of pipeline for 24 hours (default is 12 hours)</summary>
+
+```yaml
+- job: tests
+  trigger: pull_request
+  identifier: "performance"
+  targets:
+    - centos-stream-9-x86_64
+  skip_build: true
+  manual_trigger: true
+  labels:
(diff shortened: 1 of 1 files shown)


docs/configuration/examples.md
```diff
@@ -431,6 +431,30 @@ jobs:
 
 </details>
 
+<details>
+  <summary>Extending timeout of pipeline for 24 hours (default is 12 hours)</summary>
```
lbarcziova, line 435:
```suggestion
  <summary>Extending timeout of a test execution to 24 hours (default is 12 hours)</summary>
```

just so that it doesn't clash with Packit's pipeline terminology (see https://dashboard.packit.dev/pipelines)

lbarcziova (commented):
thanks again, this can be really helpful!

just one suggestion

docs/configuration/examples.md
see-quick: Yeah sure 💯 Thanks :)

lbarcziova (approved):

PR #868, opened 2024-04-08
Fix "Permission overview" heading in Fedora guide

lbarcziova:
So that it is linkable and visible in side TOC.


majamassarini (approved):
👍🏻

lbarcziova:
regate

PR #869, opened 2024-04-08
Document supported triggers for `tests` jobs

LecrisUT:
I have just copied the block from `copr_build`, should double-check it is correct, or maybe should specify that the triggers are equivalent with those in `copr_build`?


lbarcziova (approved):
thanks!!

PR #880, opened 2024-05-02
Minor sync release updates

lbarcziova:
Related to packit/packit#2297

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END

### DIFF
--- a/docs/configuration/examples.md
+++ b/docs/configuration/examples.md
@@ -71,7 +71,7 @@ fix-spec-file:
 
 </details>
 
-### Custom archive creation
+### Custom archive creation for SRPM
 <details>
   <summary>Creating archive with custom make target</summary>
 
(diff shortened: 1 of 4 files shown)


docs/configuration/examples.md
```diff
@@ -107,40 +107,47 @@ create-archive:
 
 </details>
 
-### Custom changelog generation
+### Custom archive creation for release syncing
+
+If you need to generate local archive(s) during release syncing, you can utilise e.g. `pre-sync` action
+to place the commands necessary for the creation. You also have to include the archive(s) in the list of files to
```
majamassarini, line 113: Not sure but something like the following is more clear to me.

```suggestion
If you need to generate local archive(s) during release syncing, you can customize e.g. the `pre-sync` action
with the commands necessary for the creation. You also have to include the archive(s) in the list of files to
```

docs/configuration/examples.md
```diff
@@ -107,40 +107,47 @@ create-archive:
...
+be moved to the dist-git repo so that it is then handled by Packit from there.
+
+
 <details>
-  <summary>Using changelog entry from a file</summary>
+  <summary>Creating archive using tito and syncing it to dist-git repo</summary>
```
majamassarini, line 118:
```suggestion
  <summary>Creating an archive using tito and syncing it to dist-git repo</summary>
```

docs/configuration/examples.md
```diff
@@ -107,40 +107,47 @@ create-archive:
...
+    dest: .
 ```
 
 </details>
 
+
+### Custom changelog generation
 <details>
-  <summary>Using `git log` output as a changelog entry </summary>
+  <summary>Using changelog entry from a file</summary>
```
majamassarini, line 136: Just because the title below has this structure and I think it is good.

```suggestion
  <summary>Using a file as a changelog entry</summary>
```

majamassarini (approved):

docs/configuration/examples.md
nforro: I agree with Maja, but I would suggest this:
```suggestion
  <summary>Using file content as a changelog entry</summary>
```

lbarcziova: thanks for the suggestions!

PR #889, opened 2024-05-28
Add simple tutorial for build in-mock

majamassarini:


docs/cli/build/in-mock.md
```diff
@@ -8,6 +8,21 @@ sidebar_position: 25
...
+If running `packit build in-mock` you got the following error, probably your configuration is broken.
+```
+ERROR: Non-existing Mock config 'default'
+```
+
+And you can use the `--root` option to solve the problem (or create/fix the default mock configuration).
+```
+packit build in-mock --root=fedora-40-x86_64
+```
+
```
nforro, line 25: I think something like this would be enough:

```diff
+:::tip
+If you get `ERROR: Non-existing Mock config 'default'`, make sure the `/etc/mock/default.cfg` file
+is valid or use +the `-r` or `--root` option to specify a chroot, for example:
+
+```
+packit build in-mock --root=fedora-40-x86_64
+```
+:::
```

nforro (approved):

Issue #899, opened 2024-07-01
Market OpenScanHub integration prototype


lbarcziova:
Followup of packit/packit-service#2454.

TODO:
- [x] document the new config option
- [x] describe in detail how the differential scan is done (where we take the base build from), link the docs of setting up Copr builds for commit (and maybe using custom Copr repos)
- [x] call for feedback, emphasize this is a prototype
- [x] create a short blogpost?

---
*Part of https://github.com/packit/packit-service/issues/2107*

PR #909, opened 2024-07-24
Openscanhub prototype docs

lbarcziova:
Related to #899
Related to packit/packit-service#2472

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -749,6 +749,13 @@ We provide the following variables:
 (*string*)  Specifies the SIG (special interest group). This option is used
 during cloning of the dist-git repos that are maintained in the SIG namespaces.
 
+#### sync_test_job_statuses_with_builds
+
+(*bool*) When set to `true` (the default), test job statuses are updated together with their corresponding
+build job statuses. When set to `false`, while test jobs are waiting for their corresponding build jobs
+to finish, their statuses remain in pending state and only build job statuses are updated, for example with
+`SRPM build is in progress...` or `Starting RPM build...`.
+
 #### osh_diff_scan_after_copr_build
 
 (*bool*) Whether to run a differential scan in [OpenScanHub](https://openscanhub.fedoraproject.org/) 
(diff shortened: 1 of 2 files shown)


docs/configuration/index.md
```diff
@@ -749,6 +749,30 @@ We provide the following variables:
 (*string*)  Specifies the SIG (special interest group). This option is used
 during cloning of the dist-git repos that are maintained in the SIG namespaces.
 
+#### sync_test_job_statuses_with_builds
```
majamassarini, line 752: I would write down when an user needs to put this key to `false`. I think it is not straightforward guessing it.

nforro (approved):

docs/configuration/index.md
```diff
@@ -749,6 +749,30 @@ We provide the following variables:
...
+also a Copr build job with `trigger: commit` present in the configuration.
+Additionally, the target branch of the pull request must match the branch specified in 
+the configuration for that job. For more details on setting up such a job, refer to
+[this documentation](/docs/configuration/upstream/copr_build.md).
+
+:::info
+This functionality is an initial prototype and both configuration and functionality
+will likely change in the future.
+We will be happy for any feedback regarding it, [please contact us](/#contact)!
+:::
```
lachmanfrantisek, line 774: :+1:

docs/configuration/index.md
```diff
@@ -749,6 +749,30 @@ We provide the following variables:
 (*string*)  Specifies the SIG (special interest group). This option is used
 during cloning of the dist-git repos that are maintained in the SIG namespaces.
 
+#### sync_test_job_statuses_with_builds
+
+(*bool*) When set to `true` (the default), test job statuses are updated together with their corresponding
+build job statuses. When set to `false`, while test jobs are waiting for their corresponding build jobs
+to finish, their statuses remain in pending state and only build job statuses are updated, for example with
+`SRPM build is in progress...` or `Starting RPM build...`.
```
majamassarini, line 757: Shouldn't we write down when an user needs to put this key to `false`. I think it is not straightforward guessing it. I see you just moved this doc...
Still I will add something like the following...
```suggestion
`SRPM build is in progress...` or `Starting RPM build...`.
You may want to disable the statuses update when you are experiencing problems with the number of Github API calls in your organization. This will lower the calls made by the Packit app.
```

majamassarini (approved):

docs/configuration/index.md
majamassarini: too late 😅

nforro: @majamassarini We can extend the description later if we see that it really has an effect and how big the effect is.

lbarcziova: @majamassarini yes that's a good point, for context it would be nice to mention it

PR #910, opened 2024-07-25
Add notes about ignoring artifacts

LecrisUT:
Related to https://src.fedoraproject.org/rpms/jowl/pull-request/17

For some reason in that PR there were many artifacts that were committed, the reason for that is still a mystery, maybe a bug where `/tmp/sandcastle/dist-git` still had artifacts from a previous run.

This PR tries to clarify more how the files are added to dist-git and encourage the usage of `.gitignore` to exclude artifacts. Technically `fedpkg new-sources` does take care of some of these artifacts, but things can break in weird ways as we saw there.

### DIFF
--- a/docs/configuration/examples.md
+++ b/docs/configuration/examples.md
@@ -111,7 +111,8 @@ create-archive:
 
 If you need to generate local archive(s) during release syncing, you can utilise e.g. `pre-sync` action
 to place the commands necessary for the creation. You also have to include the archive(s) in the list of files to
-be moved to the dist-git repo so that it is then handled by Packit from there.
+be moved to the dist-git repo so that it is then handled by Packit from there. **Note:** Beware of how Packit uploads
+files to lookaside cache or adds them to dist-git (see [this note][files_to_sync_info] for more details).
 
 
 <details>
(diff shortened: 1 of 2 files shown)


docs/configuration/examples.md
```diff
@@ -111,7 +111,8 @@ create-archive:
 
 If you need to generate local archive(s) during release syncing, you can utilise e.g. `pre-sync` action
 to place the commands necessary for the creation. You also have to include the archive(s) in the list of files to
-be moved to the dist-git repo so that it is then handled by Packit from there.
+be moved to the dist-git repo so that it is then handled by Packit from there. **Note:** make sure the archive(s)
+are excluded by the downstream `.gitignore`.
```
lbarcziova, line 115: I'm not sure it makes sense to include it here as I would say it is the intention for the files to be included if they are specified in the `files_to_sync`, WDYT?

LecrisUT: Hmm, kinda, but this is the only example for these. Let me try some rewording for this to be more specific about lookaside uploads

lbarcziova (approved):
thanks!

PR #923, opened 2024-08-28
Document how to release multiple packages in one update

nforro:
Related to https://github.com/packit/packit/issues/1870.

### DIFF
--- a/docs/configuration/downstream/bodhi_update.md
+++ b/docs/configuration/downstream/bodhi_update.md
@@ -39,6 +39,9 @@ Packit dist-git configuration.
 
 * **commit** - Packit uses the original action as a config trigger, so you need to use `commit` as a trigger.
   The real trigger is a successful Koji build (that was triggered from a commit).
+* **koji_build** - reacts to tagging of a Koji build into the configured sidetag group
+  More info about releasing multiple packages together using sidetags can be found
+  [here](/docs/fedora-releases-guide/releasing-multiple-packages.md).
 
 ## Required parameters
 
@@ -55,6 +58,13 @@ Packit dist-git configuration.
      (don't forget to put the group name in quotes - a token in YAML is not allowed to start with `@`)
    - `all_admins` alias - allowing all users with admin access to the dist-git repo
    - `all_committers` alias - allowing all users with commit access to the dist-git repo
+* **sidetag_group** - name of a sidetag group this Bodhi update should be created from.
+  The name has to be unique across all projects that use Packit, so don't hesitate to be descriptive.
+  A good convention is to use `${package1}-${package2}...${packageN}-updates`.
+* **dependencies** - list of downstream package names that are required to be tagged
+  into the configured sidetag group for this job to be triggered. For example, if you want to include
+  a `koji_build` job with the same sidetag group configured in package `foo` in this update,
+  you need to list `foo` here.
 
  
 ## Example
--- a/docs/configuration/downstream/koji_build.md
+++ b/docs/configuration/downstream/koji_build.md
@@ -50,10 +50,22 @@ Packit dist-git configuration.
 ## Supported triggers
 
 * **commit** - reacts to new commits to the specified branch (in dist-git)
+* **koji_build** - reacts to tagging of a Koji build into the configured sidetag group
+  More info about releasing multiple packages together using sidetags can be found
+  [here](/docs/fedora-releases-guide/releasing-multiple-packages.md).
+
+:::tip You can use both triggers simultaneously
+
+```yaml
+- job: koji_build
+  trigger: commit | koji_build
+```
+
+:::
 
 ## Required parameters
 
-* **dist_git_branches** - the name of the dist-git branch we want to build for when using **commit** trigger.
+* **dist_git_branches** - the name of the dist-git branch we want to build for.
   You can also use the [aliases provided by Packit](/docs/configuration#aliases)
   to not need to change the config file when the new system version is released.
 
@@ -77,8 +89,21 @@ Packit dist-git configuration.
      (don't forget to put the group name in quotes - a token in YAML is not allowed to start with `@`)
    - `all_admins` alias - allowing all users with admin access to the dist-git repo
    - `all_committers` alias - allowing all users with commit access to the dist-git repo
+* **sidetag_group** - name of a sidetag group this Koji build should be tagged into.
+  The name has to be unique across all projects that use Packit, so don't hesitate to be descriptive.
+  A good convention is to use `${package1}-${package2}...${packageN}-updates`.
+  If specified, after a build finishes it is tagged into a sidetag (that is created if it doesn't already exist)
+  corresponding to the dist-git branch of the build.
+* **dependencies** - list of downstream package names that are required to be tagged
+  into the configured sidetag group for this job to be triggered. For example, if you want to ensure
+  that a `koji_build` job with the same sidetag group configured in package `foo` is run first,
+  you need to list `foo` here.
+* **dependents** - list of downstream package names that should react to tagging of this build
+  into the configured sidetag group. For example, if a `koji_build` job with the same sidetag group
+  configured in package `bar` is supposed to be triggered after this build finishes, you need to
+  list `bar` here.
 * **require.label** - you can specify labels that have to be present/absent on a pull request
-in order to trigger the build when it is merged. See configuration details [here](/docs/configuration#require).
+  in order to trigger the build when it is merged. See configuration details [here](/docs/configuration#require).
 
 
 ## Processing of dist-git events from Pagure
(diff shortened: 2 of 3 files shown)


docs/configuration/downstream/bodhi_update.md
```diff
@@ -39,6 +39,7 @@ Packit dist-git configuration.
 
 * **commit** - Packit uses the original action as a config trigger, so you need to use `commit` as a trigger.
   The real trigger is a successful Koji build (that was triggered from a commit).
+* **koji_build** - reacts to tagging of a Koji build into the configured sidetag group
```
lachmanfrantisek, line 42: What do you think about linking the blog post from related places like this?

```suggestion
* **koji_build** - reacts to tagging of a Koji build into the configured sidetag group. (More info about releasing multiple packages can be found [here](/docs/fedora-releases-guide/releasing-multiple-packages.md))
```

docs/configuration/downstream/koji_build.md
```diff
@@ -77,8 +87,16 @@ Packit dist-git configuration.
      (don't forget to put the group name in quotes - a token in YAML is not allowed to start with `@`)
    - `all_admins` alias - allowing all users with admin access to the dist-git repo
    - `all_committers` alias - allowing all users with commit access to the dist-git repo
+* **sidetag_group** - name of a sidetag group this Koji build should be tagged into.
+  The name has to be unique across all projects that use Packit, so don't hesitate to be descriptive.
```
lachmanfrantisek, line 91: Ah, that's interesting. A few questions unrelated to docs:
* Can we provide a list of Packit's used side-tags to check / get inspired? It's being removed, right?
* Can this be somehow misused?
* Can't we provide a default or suggestions? (Like `{package1}-{package2}-releases`)?

docs/fedora-releases-guide/releasing-multiple-packages.md
```diff
@@ -0,0 +1,198 @@
+---
+title: Releasing multiple packages in a single update
+sidebar_position: 7
+---
+# Releasing multiple packages in a single update
```
lachmanfrantisek, line 5: Very nicely written and explained! I like the 3 scenario structure and that we can have a link for each.

docs/fedora-releases-guide/releasing-multiple-packages.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Skipping release of some packages
+
+But what if we only want to release `bar` and not `foo`? We can tag the latest stable build of `foo`
+to the common sidetag to satisfy the dependency and trigger the jobs in `bar`. We can do that
+by commenting `/packit koji-tag --all-branches` in any dist-git pull request of `foo`. By omitting
+the `--all-branches` argument we can make the command apply only to the target branch
+of the pull request.
+
+## Full example
```
lachmanfrantisek, line 117: This really helps with the understanding...

lachmanfrantisek (approved):
Thanks!

docs/configuration/downstream/koji_build.md
```diff
@@ -77,8 +87,16 @@ Packit dist-git configuration.
...
+  If specified, after a build finishes it is tagged into a sidetag (that is created if it doesn't already exist)
+  corresponding to the dist-git branch of the build.
+* **dependencies** - list of downstream package names that are required to be tagged
+  into the configured sidetag group for this job to be triggered.
+* **dependents** - list of downstream package names that should react to tagging of this build
+  into the configured sidetag group.
```
majamassarini, line 97: I think this are nicely written, but to me they are really hard to grasp. Can we add an example?
```suggestion
* **dependencies** - list of downstream package names that are required to be tagged
  into the configured sidetag group for this job to be triggered. As an example the `packit` package depends  on the `python-specfile` package. Hence in `packit.yaml` inside `packit` project you find `python-specfile` package listed in the `dependencies`.
* **dependents** - list of downstream package names that should react to tagging of this build
  into the configured sidetag group. As an example the `packit` package depends  on the `python-specfile` package. Hence in `packit.yaml` inside `specfile` project you find `packit` package listed in the `dependents`.
```

docs/fedora-releases-guide/releasing-multiple-packages.md
```diff
@@ -0,0 +1,198 @@
...
+  - job: bodhi_update
+    trigger: koji_build
+    sidetag_group: foo-bar-releases
+    dependencies:
+      - foo
+    dist_git_branches:
+      - fedora-all
+```
+
+Note there is only one `bodhi_update` job for both packages, we chose to configure it in `bar`.
```
majamassarini, line 51:
```suggestion
Note that there is only one `bodhi_update` job for both packages, we chose to configure it in `bar`.
```
Not sure, but it sounds to me more correct.

docs/fedora-releases-guide/releasing-multiple-packages.md
```diff
@@ -0,0 +1,198 @@
...
+and, vice versa, reference `foo` from `dependencies` of the `bodhi_update` job in `bar`.
+The `sidetag_group` option, having the same value in all jobs, defines the sidetags Koji builds
+will be tagged into and Bodhi update will be created from. It doesn't matter which package
+is built first, only after builds of both of them are tagged into the sidetag a Bodhi update
+is created and the sidetag removed.
+
+## Ensuring build order
+
+Let's have the same two packages as before, but this time `bar` depends on `foo`, meaning that
+a build of `bar` can fail if it is build before `foo`. So we have to make sure that `foo` is always
```
majamassarini, line 62:
```suggestion
a build of `bar` can fail if it is built before `foo`. So we have to make sure that `foo` is always
```

docs/fedora-releases-guide/releasing-multiple-packages.md
```diff
@@ -0,0 +1,198 @@
...
+
+Note that config of `foo` hasn't changed at all. The `koji_build` job in `bar` however now has
+two triggers - it will be triggered also when a build of `foo` lands in the common sidetag.
+It also has `foo` in `dependencies`, so if it is triggred via commit while `foo` is not yet
+in the sidetag, it will be skipped. As before, once builds of both packages are in the sidetag,
+a Bodhi update is created and the sidetag removed.
+
+## Skipping release of some packages
+
+But what if we only want to release `bar` and not `foo`? We can tag the latest stable build of `foo`
```
majamassarini, line 111:
```suggestion
What if we only want to release `bar` and not `foo`? We can tag the latest stable build of `foo`
```

I would not start a new paragraph with *but*.

majamassarini (approved):
🚀

docs/configuration/downstream/koji_build.md
nforro: > * Can we provide a list of Packit's used side-tags to check / get inspired? It's being removed, right?

Sidetags or sidetag groups? I assume you mean sidetag groups, as users (Packit included) have no control over sidetag naming. Well, we could provide a list, but there is only one group so far, `packit-specfile-ogr-releases`. Sidetag groups are not removed (they could be removed as part of https://github.com/packit/packit-service/issues/2412, but that's not implemented yet), sidetags are.

> * Can this be somehow misused?

I suppose so, you could tag "unwanted" builds into someone else's sidetags or untag someone else's builds, but you can do the same with `koji tag-build/untag-build` - there are no sidetag permissions, any packager is allowed tag/untag anything into/from a sidetag.

> * Can't we provide a default or suggestions? (Like `{package1}-{package2}-releases`)?

Sure, if you think this format/convention makes sense then I'll include it as a suggestion.

docs/fedora-releases-guide/releasing-multiple-packages.md
nforro: I was trying to avoid something like `That's all cool and dandy, but what if ...` :sweat_smile:

docs/fedora-releases-guide/releasing-multiple-packages.md
nforro: Both are correct. But since there is another occurence of `Note that` below, let's make it consistent.

docs/configuration/downstream/koji_build.md
nforro: What about links to the guide/examples as Franta suggests?

majamassarini: The links are useful, and I would take them, but personally if I small example can help I prefer having it instead of having to read a long guide.

nforro: I've added examples, what do you think?

docs/configuration/downstream/koji_build.md
nforro: > Sure, if you think this format/convention makes sense then I'll include it as a suggestion.

I've added `${package1}-${package2}...${packageN}-updates`, I think `updates` is more fitting from downstream maintainer PoV, but I can change it to `releases`.

PR #935, opened 2024-09-18
Two doc fixes

lachmanfrantisek:
I've misinterpreted the Docusaurus docs that the image-like markdown syntax should work with youtube videos but it does not. (They suggest creating your own plugin to do this.)


lbarcziova (approved):
I just noticed there is also missing `)` in the first sentence, could you please fix that directly?

PR #943, opened 2024-10-01
Add benefits of using packit to faq

### LINKED ISSUE #794: Document aditional benefits of Packit compared to usage of external services directly
Packit delegates most of the work to external services and here and there, we are asked to provide the benefits of Packit compared to using those services directly.

Let's provide a summary of these benefits on our documentation page.

This info needs to be linkable and can be either put on the homepage or maybe in the FAQs. (Other suggestions welcome.) Alternatively, we can provide this info for each job separately, but most probably, the benefits are generic.

kapr200:
Adds the Documentation about aditional benefits of Packit compared to usage of external services directly.

Fixes #794

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -157,3 +157,14 @@ You can read more about this transition [here](/posts/copr-srpms).
 ## A command failed in packit-service: how do I reproduce it locally?
 
 Please read [this document on how to reproduce locally](reproduce-locally).
+
+## Why you should use Packit instead of external devices directly?
+
+- Packit uses various services Testing Farm, Copr, Koji, Bodhi…
+- These services can be used without Packit, but Packit integrates them in a
+  more user-friendly way reducing the manual labor that would be required from
+  the contributor or maintainer. For example:
+  - the build in Copr is created automatically when a new version is released
+  - RPMs built by Copr are specifically passed to the Testing Farm for testing
+  - RPM builds are verified against multiple releases (and also different OSs)
+- The user does not have to wait and look for the results but just interprets them.


docs/faq.md
```diff
@@ -157,3 +157,9 @@ You can read more about this transition [here](/posts/copr-srpms).
 ## A command failed in packit-service: how do I reproduce it locally?
 
 Please read [this document on how to reproduce locally](reproduce-locally).
+
+## Unlocking Efficiency: The Advantages of Using Packit for Seamless Service Integration
+
+- Packit uses various services Testing Farm, Copr, Koji, Bodhi…
+- These services can be used without Packit, but Packit intergrates their continuity (the build in Copr is created automatically when a new version is released).
```
majamassarini, line 164:
```suggestion
- These services can be used without Packit, but Packit integrates them (the build in Copr is created automatically when a new version is released).
```
My english is bad, but the sentence above sounds to me better in the way I propose, but feel free to ignore my suggestion (just apply the typo fix :-))

majamassarini (approved):

docs/faq.md
```diff
@@ -157,3 +157,9 @@ You can read more about this transition [here](/posts/copr-srpms).
 ## A command failed in packit-service: how do I reproduce it locally?
 
 Please read [this document on how to reproduce locally](reproduce-locally).
+
+## Unlocking Efficiency: The Advantages of Using Packit for Seamless Service Integration
```
lachmanfrantisek, line 161: Since this is a `faq` part, I would try to convert this into a question.

docs/faq.md
mfocko: Sounds better to me :eyes:

docs/faq.md
lachmanfrantisek: Maybe only me, but `Seamless Service Integration` seems a bit too buzzwordy to me...;) I think we don't need to create much hype in Frequently Asked Questions..;)

docs/faq.md
lachmanfrantisek: I would also maybe explicitly emphasise that in some workflows (mainly Fedora maintenance automation), we are chaining the tasks run on these services so people do not need to be busy waiting.

lachmanfrantisek (commented):
Thank you! This is important and not easy to write in a few simple sentences I must say.

docs/faq.md
```diff
@@ -157,3 +157,9 @@ You can read more about this transition [here](/posts/copr-srpms).
...
+
+- Packit uses various services Testing Farm, Copr, Koji, Bodhi…
+- These services can be used without Packit, but Packit intergrates their continuity (the build in Copr is created automatically when a new version is released).
+- The user does not have to wait for the results and just interprets them.
```
lbarcziova, line 165: here I would emphasize also that the results can be accessed from one place (though indirectly), the forge/dashboard

docs/faq.md
lbarcziova: +1, I would in general extend this a little bit more, e.g. with how we take care of passing the built NVRs of RPM builds to Testing Farm

mfocko (approved):

docs/faq.md
```diff
@@ -157,3 +157,9 @@ You can read more about this transition [here](/posts/copr-srpms).
 ## A command failed in packit-service: how do I reproduce it locally?
 
 Please read [this document on how to reproduce locally](reproduce-locally).
+
+## Why you should use Packit instead of external devices directly?
+
+- Packit uses various services Testing Farm, Copr, Koji, Bodhi…
+- These services can be used without Packit, but Packit integrates them (the build in Copr is created automatically when a new version is released).
```
mfocko, line 164:
```suggestion
- These services can be used without Packit, but Packit integrates them in a
  more user-friendly way reducing the manual labor that would be required from
  the contributor or maintainer. For example:
  - the build in Copr is created automatically when a new version is released
  - RPMs built by Copr are specifically passed to the Testing Farm for testing
  - RPM builds are verified against multiple releases (and also different OSs)
```

@lbarcziova WDYT?

mfocko (approved):

docs/faq.md
lbarcziova: LGTM, much better, thanks!

PR #971, opened 2024-11-13
Document the `_` placeholder

LecrisUT:
Related to packit/packit/pull/2378

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -16,6 +16,13 @@ You should place the file in the root of your upstream repo. Packit accepts thes
 
 Both Packit Service and Packit CLI use this configuration file.
 
+:::tip Writing complex configurations
+
+See the [Packit Service jobs configuration](/docs/configuration/jobs.md) section for some tips on how to configure
+complex config files.
+
+:::
+
 ## Top-level keys
 
 Here you can see a list of options that can be defined at the top-level and
--- a/docs/configuration/jobs.md
+++ b/docs/configuration/jobs.md
@@ -78,9 +78,9 @@ jobs:
   branch: main
 ```
 
-Do not forget that you can also use the top-level keys to store common keys like
-`targets` and `actions` and only override them as needed in the `jobs` or
-`packages` section.
+Do not forget that you can also define common Packit config options at the
+top level (such as `targets` or `actions`) and only override them when
+needed in the `jobs` or `packages` section.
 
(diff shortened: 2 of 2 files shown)


docs/configuration/index.md
```diff
@@ -16,6 +16,12 @@ You should place the file in the root of your upstream repo. Packit accepts thes
 
 Both Packit Service and Packit CLI use this configuration file.
 
+:::tip Writing complex configurations
+
+See [](/docs/configuration/jobs.md) for some tips on how to configure complex config files.
```
LecrisUT, line 21: Welp, docusaurus does not seem to support auto link :/

docs/configuration/jobs.md
```diff
@@ -78,13 +78,43 @@ jobs:
...
 ```
 
-Note that in case of more complex scenarios, it can be useful to create also
-a template job, that should not be triggered in Packit at all.
-This is typically useful when creating a number of tests which needs various
-parameters, but many parameters stay same. In such a case set the 
-`trigger` to `ignore` in the template job and use the expected trigger in real jobs.
+Do not forget that you can also use the top-level keys to store common keys like
+`targets` and `actions` and only override them as needed in the `jobs` or
+`packages` section.
```
mfocko, line 83: I think I know what you want to say here, but it's not doing a great job. You can't use top-level keys (except for the `_` that you allowed in the other PR), because we validate the schema.

LecrisUT: I'm trying to say that instead of having a 
```yaml
jobs:
  - &copr_build
    job: copr_build
    trigger: pull_request
    targets: &targets
      - fedora-all
      - epel-all
  - &tests
    job: tests
    trigger: pull_request
    targets: *targets
```
we should remember that we can define `targets` at the top-level
```yaml
targets:
  - fedora-all
  - epel-all
jobs:
  - &copr_build
    job: copr_build
    trigger: pull_request
  - &tests
    job: tests
    trigger: pull_request
```

That would be allowed by the schema validation. Any ideas on how to rephrase it?

mfocko:
```suggestion
Do not forget that you can also define common Packit config options at the
top level (such as `targets` and `actions`) and only override them when
needed in the `jobs` or `packages` section.
```

maybe?

LecrisUT: Sounds good to me. I'll leave you to apply it if you have some other edits you want to stack up.

mfocko (approved):

PR #986, opened 2025-01-08
Add a dedicated section for tests env vars

thrix:
It is better for referencing, to have a dedicated and linkable section for these.


thrix:
@lbarcziova @mfocko howdy, could we include this pls? It would make linking to this important section a lot easier.

PR #1003, opened 2025-02-17
Add blog post for week 7 in 2025

lbarcziova:


weekly/2025/week-7.md
```diff
@@ -0,0 +1,15 @@
...
+authors: lbarczio
+tags:
+  - 2025-February
+  - 2025
+  - February
+---
+
+## Week 7 (February 11th – February 17th)
+
+- Packit now reloads spec files after all of `actions` that might modify them during release syncing.
```
nforro, line 13:
```suggestion
- Packit now reloads spec files after all `actions` that might modify them during release syncing.
```

nforro (approved):
Technically, the reloading happens also for all "upstream" actions as well, but we don't have to mention it as it doesn't fix any real use case.

PR #1010, opened 2025-03-17
Document EPEL minor version aliases

nforro:
Related to https://github.com/packit/packit/issues/2438.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -828,6 +828,46 @@ The information about releases [is retrieved from Bodhi](https://bodhi.fedorapro
 cache and required availability on Copr, it might take a while to get the
 newest state.
 
+### EPEL minor version aliases
+
+Starting with EPEL 10, there are [different dist-git branches](https://docs.fedoraproject.org/en-US/epel/branches/#_epel_10)
+and Koji targets for currently active minor versions:
+
+* The main/default branch/target called `epelX` (e.g. `epel10`) that represents
+  the latest (leading) minor release and corresponds to CentOS Stream X.
+* The `epelX.Y` (e.g. `epel10.0`) branch/target that represents
+  the previous (trailing) minor release and corresponds to RHEL X.Y.
+
+:::info
+
+See the [EPEL 10 proposal](https://discussion.fedoraproject.org/t/epel-10-proposal)
+for more information, FAQ and discussion about this change and don't hesitate to provide
+your feedback through proper channels, for instance
+the [epel-devel](https://lists.fedoraproject.org/archives/list/epel-devel@lists.fedoraproject.org/)
+mailing list.
+
+:::
+
+To give you flexibility in what distributions you want to target with your EPEL builds,
+Packit provides these aliases:
+
+* `epel-X` — the current/leading minor release (e.g. `epel10`)
+* `epel-X-branched` — the branched/trailing minor release (e.g. `epel10.0`)
+* `epel-X-all` — both minor releases (e.g. `epel10`, `epel10.0`)
+
+::: note
+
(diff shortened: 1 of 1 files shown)


docs/configuration/index.md
```diff
@@ -828,6 +828,40 @@ The information about releases [is retrieved from Bodhi](https://bodhi.fedorapro
 cache and required availability on Copr, it might take a while to get the
 newest state.
 
+### EPEL minor version aliases
+
+Starting with EPEL 10, there are [different dist-git branches](https://docs.fedoraproject.org/en-US/epel/branches/#_epel_10)
+and Koji targets for currently active minor versions. The main/default branch/target
+called `epelX` (e.g. `epel10`) that represents the latest (leading) minor release
+and corresponds to CentOS Stream X and the `epelX.Y` (e.g. `epel10.0`) branch/target
+that represents the previous (trailing) minor release and corresponds to RHEL X.Y.
```
majamassarini, line 837:
```suggestion
and Koji targets for currently active minor versions. 

The main/default branch/target called `epelX` (e.g. `epel10`)  represents the latest (leading) minor release
and corresponds to CentOS Stream X. 

The `epelX.Y` (e.g. `epel10.0`) branch/target represents the previous (trailing) minor release and corresponds to RHEL X.Y.
```

This is though to read. I would try to spilt it up.

majamassarini (approved):

docs/configuration/index.md
nforro: What I could do is to convert the second sentence into a 2-item list, that would help readability, but at the same time it would feel a bit weird to me.

majamassarini: I was lost after the first `that` sentence, but maybe it is just me. Feel free to ignore my suggestion. Having a list can surely improve the readability.

lbarcziova (commented):
can we also include some more explicit call for feedback and mention we may change things depending on that?

docs/configuration/index.md
```diff
@@ -828,6 +828,43 @@ The information about releases [is retrieved from Bodhi](https://bodhi.fedorapro
...
 
+### EPEL minor version aliases
+
+Starting with EPEL 10, there are [different dist-git branches](https://docs.fedoraproject.org/en-US/epel/branches/#_epel_10)
+and Koji targets for currently active minor versions:
+
+* The main/default branch/target called `epelX` (e.g. `epel10`) that represents
+  the latest (leading) minor release and corresponds to CentOS Stream X.
+* The `epelX.Y` (e.g. `epel10.0`) branch/target that represents
+  the previous (trailing) minor release and corresponds to RHEL X.Y.
```
majamassarini, line 839: 🙏🏻

lbarcziova:
@lbarcziova I meant not only on Packit side, but the whole Copr + fedora-distro-aliases workflows, as we discussed on standup.

nforro:
> @lbarcziova I meant not only on Packit side, but the whole Copr + fedora-distro-aliases workflows, as we discussed on standup.

I suppose that was a bit of a misunderstanding, it was about feedback to the (already implemented) EPEL proposal and the fact that maintainers have to maintain two EPEL branches if they want their builds to land in a released RHEL minor version and not only in the future one. There isn't much Packit or Copr can do about this, there will be a pair of targets for EPEL 10, we can only affect the naming and whether we want to somehow motivate users to build only in the leading one (but that concerns mainly downstream jobs).
To be honest I'm not sure where to point users to voice their opinions about this, apart from the epel-devel mailing list.

lbarcziova (approved):
Ok, thanks for summarising, I would say mentioning the epel-devel mailing list won't hurt, but will leave it up to you. LGTM otherwise :)

PR #1012, opened 2025-03-18
Fix silly typos

nforro:
Related to https://github.com/packit/packit.dev/pull/1010.


majamassarini (approved):

mfocko:
> I wish we had a pre-commit hook to check for these types of errors 😅

There should be one, but I'm afraid that it won't play very well with the variety of names :D

nforro:
> There should be one, but I'm afraid that it won't play very well with the variety of names :D
> 
> https://github.com/crate-ci/typos

I meant more like some checker for this specific syntax of Markdown admonitions - after all these are more syntax errors than typos :slightly_smiling_face:

PR #1024, opened 2025-04-07
add documentation for osh options

kayodegigz:
Related to packit/packit#2570


mfocko (approved):
Thanks a lot, I'll do some minor stylistic updates, it will be a bit faster that way

PR #1046, opened 2025-05-30
Add missing '.' to release suffix pattern

mulkieran:


mfocko (approved):

docs/configuration/index.md
```diff
@@ -218,7 +218,7 @@ Example usage:
 
 When unset, default release suffix that is generated by Packit is following:
 
-    {original_release_number}.{current_time}.{sanitized_current_branch}{git_desc_suffix}
+    {original_release_number}.{current_time}.{sanitized_current_branch}.{git_desc_suffix}
```
mfocko, line 221: It appears like that, but it's a consequence of the `git_desc_suffix` being prefixed with the `.`, so in the template, the dot is actually not present…

mulkieran: Aaah!

mulkieran:
I think you're better off saying something like:

"If ```git describe``` does not provide a value for ```git_desc_suffix``` that part of the release suffix is omitted."

That'll make clear it's a special field at least, without going into details about Packit formatting choices that may change.

PR #1081, opened 2026-01-08
Document sandbox limits

nforro:
Fixes https://github.com/packit/packit-service/issues/2917.

### DIFF
--- a/docs/configuration/actions.md
+++ b/docs/configuration/actions.md
@@ -24,8 +24,12 @@ bash -c "my fancy $command | grep success"
 > It's important to quote the content of the `-c` option so shell interprets it
 > correctly as a single option input.
 
-All actions are also executed inside Packit Service. The service
-creates a new sandbox environment where the command is run.
+SRPM creation actions run in Copr while release syncing actions are executed
+in a sandbox environment inside Packit Service. This environment has limited resources.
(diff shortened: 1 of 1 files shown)


docs/configuration/actions.md
```diff
@@ -24,8 +24,11 @@ bash -c "my fancy $command | grep success"
 > It's important to quote the content of the `-c` option so shell interprets it
 > correctly as a single option input.
 
-All actions are also executed inside Packit Service. The service
-creates a new sandbox environment where the command is run.
+SRPM creation actions run in Copr while release syncing actions are executed
+in a sandbox environment inside Packit Service. This environment has limited resources.
+Your actions shouldn't consume more than 768MiB of memory and use more than 3GiB of disk space.
+Be aware that cloned upstream and dist-git repos also count towards this limit.
+It is recommended to place any temporary artifacts in a subdirectory of `/tmp`, as there is no limit.
```
nforro: I'm a bit unsure about this, because the limited volume with cloned git repos is mounted at `/tmp/sandcastle`, which is technically also a subdirectory of `/tmp`.

majamassarini: I am not sure either.
`oc describe limitrange -n packit--prod-sandbox`
Shows that the default limit for the storage in `packit--prod-sandbox` is 2Gi, I expect it to apply outside the mount point of our volume claim.

PR #1136, opened 2026-05-05
Update versions in alias examples

nforro:
While not technically wrong, outdated Fedora and EPEL versions in the examples can be confusing to users.


majamassarini (approved):
🙏🏻

PR #1138, opened 2026-05-18
fix(tf): mention allowlisting of repos on staging

mfocko:
Outcome from the Friday's troubleshooting of running internal TF from the GitLab.com via Packit Stagin with @modehnal.


lbarcziova (approved):
thanks!

PR #1142, opened 2026-05-26
Add guide for tracking Source files in dist-git

majamassarini:
Add a how-to that shows how to use the post-modifications action (or pre-sync for URL-based sources) together with files_to_sync to keep small extra Source files in dist-git's VCS rather than uploading them to the lookaside cache on the first sync.

Link to the new guide from the existing note in the configuration index and from the actions reference.

Fixes packit/packit#2365

Assisted-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/docs/configuration/actions.md
+++ b/docs/configuration/actions.md
@@ -170,12 +170,13 @@ You can, of course, provide *just* the commit title.
 
 ### `post-modifications`
 
-This action runs in the release-synchronization workflow after the spec file
-has been modified and remote sources have been downloaded into dist-git, but
-*before* local `Source` files are collected for the lookaside-cache upload.
-You can use it to stage extra files (e.g. systemd units, configuration templates) into
-dist-git's Git index so that Packit treats them as VCS-tracked and skips them
-when uploading to the lookaside cache. A complete walkthrough is in
+During release synchronization (`propose_downstream` / `pull_from_upstream`),
+this action runs after the spec file has been modified and remote sources have
+been downloaded into dist-git, but *before* local `Source` files are collected
+for the lookaside-cache upload. You can use this timing to stage extra files
+(e.g. systemd units, configuration templates) into dist-git's Git index so that
+Packit treats them as VCS-tracked and skips them when uploading to the
+lookaside cache. A complete walkthrough is in
 [Tracking `Source` files in dist-git](/docs/configuration/downstream/source-files-in-dist-git).
 
 ### `run-condition`
--- a/docs/configuration/actions.md
+++ b/docs/configuration/actions.md
@@ -168,6 +168,16 @@ Our internal API differentiates between the commit title and description, so we
 have an additional requirement on separating the commit title by an empty line.
 You can, of course, provide *just* the commit title.
 
+### `post-modifications`
+
+This action runs in the release-synchronization workflow after the spec file
+has been modified and remote sources have been downloaded into dist-git, but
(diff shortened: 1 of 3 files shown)


docs/configuration/downstream/source-files-in-dist-git.md
```diff
@@ -0,0 +1,139 @@
...
+files_to_sync:
+  - pkg.spec
+  - pkg.service
+  - pkg.socket
+  - pkg.sysusers
+
+actions:
+  post-modifications:
+    - bash -c 'cd "${PACKIT_DOWNSTREAM_REPO}" && git add -- pkg.service pkg.socket pkg.sysusers'
+```
```
nforro: Is this meant to discourage using `git add *` or `git add .`? If yes, I would state that explicitly, becuase otherwise this whole paragraph seems redundant - everything is practically the same as for a single file.

docs/configuration/actions.md
```diff
@@ -168,6 +168,16 @@ Our internal API differentiates between the commit title and description, so we
 have an additional requirement on separating the commit title by an empty line.
 You can, of course, provide *just* the commit title.
 
+### `post-modifications`
+
+This action runs in the release-synchronization workflow after the spec file
```
nforro: Not true (or badly formulated), it runs in upstream jobs as well.

docs/configuration/actions.md
```diff
@@ -168,6 +168,16 @@ Our internal API differentiates between the commit title and description, so we
...
+has been modified and remote sources have been downloaded into dist-git, but
+*before* local `Source` files are collected for the lookaside-cache upload.
+Use it to stage extra files (e.g. systemd units, configuration templates) into
```
nforro: The primary motivation for this action was something else:

```suggestion
You can use it to stage extra files (e.g. systemd units, configuration templates) into
```

nforro (approved):
A little too verbose for my taste, but definitely better than nothing 🙂

docs/configuration/downstream/source-files-in-dist-git.md
majamassarini: The reason why the agent added this example and it isn't using `git add *` is explained.

> Using `git add -- <paths>` keeps the action robust against future
files that match dot-prefixes or unusual names:

```
  $ git add *               # so this misses .env
  $ git status -s .env
  ?? .env                   # still untracked
```

```
  $ git add -pkg.service
  unknown switch `p'
  usage: git add [<options>] [--] <pathspec>...

  $ git add -- -pkg.service     # works
```

These are really corner cases, so if you want I can remove this paragraph, but I think this is a useful tip.

docs/configuration/actions.md
majamassarini: I don't read it as it runs only in release-synchronization. I read it as it runs after the specfile has been modified during the release-synchronization. 🤔 or I don't follow you.

nforro: That's why I said badly formulated. You can read it both ways, let's use different wording.

docs/configuration/downstream/source-files-in-dist-git.md
nforro: > I think this is a useful tip.

Exactly, so I would convert it to a tip admonition and shorten it to one ot two sentences (do not use `git add .` or `git add *`, use `git add -- file1 file2` instead).

nforro (approved):

PR #1145, opened 2026-06-24
Fix invalid YAML in an example

nforro:


majamassarini (approved):
Shouldn't we wrap the numbers in double quotes also for the example below?

nforro:
> Shouldn't we wrap the numbers in double quotes also for the example below?

That's valid YAML. The question is if we want to be consistent or if we want the example to demonstrate that some expressions can be used without quotes :slightly_smiling_face:  WDYT?

majamassarini:
If it is valid then I am fine as it is.

PR #1151, opened 2026-07-16
fix(blog): add dates to the blog post

mfocko:
As I have noticed when opening the blog post about release syncing in Packit from Nikola, when you don't specify the date the blogs get assigned the date during the build.

Therefore these two blog posts are kept at the top of the posts.

Add the dates of merging to fix this issue.

Cc @TomasTomecek I don’t want to steal your limelight, but it’s time for your blog post from February, to sink down 😈

### DIFF
--- a/posts/code-review/index.md
+++ b/posts/code-review/index.md
@@ -1,6 +1,7 @@
 ---
 title: Our experience with AI-based code review
 authors: ttomecek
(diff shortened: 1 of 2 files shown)


TomasTomecek (approved):
and I was wondering why they have such a weird order, thank you!!

PR #1152, opened 2026-07-16
docs: clarify that a job trigger accepts a single value

### LINKED ISSUE #1054: Make it clear that a job `trigger` only accepts a single value
As far as I can tell, it's not mentioned anywhere. Even though all examples only show a single trigger (with the exception of sidetag-enabled downstream jobs accepting `trigger: commit | koji_build`), we should mention it explicitly.

Nitjsefnie:
Adds an explicit note to the `jobs` section of `docs/configuration/index.md`
stating that a job's `trigger` key accepts exactly one value and cannot be a
list or combination of multiple triggers.

Closes #1054

Prepared with AI assistance (Kimi K2.7 Code), human-reviewed before submission.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -76,6 +76,9 @@ files_to_sync:
 (*list of dicts*) A list of job definitions for Packit Service: see
 [Packit Service jobs configuration](configuration/jobs) for details.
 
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
LGTM, thank you!

PR #1165, opened 2026-08-13
fix(docs): add a note about bundling and `merge_pr_in_ci`

mfocko:
Fixes packit/packit#2733
Fixes PACKIT-5184

Since there is no reply on the issue, I’m adding a note to the docs and closing down the issue.


betulependule (approved):
The content is fine and looks ok when viewed in a browser. LGTM

PR #1168, opened 2026-08-26
Document the added support of `identifier` with OSH jobs

betulependule:
Related to [#3172](https://github.com/packit/packit-service/pull/3172)

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -844,6 +844,14 @@ Additionally, the target branch of the pull request must match the branch specif
 the configuration for that job. For more details on setting up such a job, refer to
 [this documentation](/docs/configuration/upstream/copr_build.md).
 
+:::caution Multiple OSH jobs
+
+If `osh_diff_scan_after_copr_build` is enabled and you have multiple Copr build jobs with the `pull_request` trigger, each will trigger its own OSH differential scan.
+To ensure each scan compares against the right base build, assign matching `identifier` values to each `pull_request`-triggered job and its corresponding `commit`-triggered job.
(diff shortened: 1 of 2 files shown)


docs/configuration/index.md
```diff
@@ -844,6 +844,10 @@ Additionally, the target branch of the pull request must match the branch specif
 the configuration for that job. For more details on setting up such a job, refer to
 [this documentation](/docs/configuration/upstream/copr_build.md).
 
+:::caution Multiple OSH jobs
+If `osh_diff_scan_after_copr_build` is enabled and you have multiple Copr build jobs with the `pull_request` trigger, each will trigger its own OSH differential scan. To ensure each scan compares against the right base build, give matching `identifier` values to each `pull_request`-triggered job and its corresponding `commit`-triggered job. If identifiers are missing or don't match, Packit falls back to the first `commit`-triggered job found for that branch — which can mean multiple scans end up comparing against the same (possibly wrong) base build.
```
nforro:
```suggestion
If `osh_diff_scan_after_copr_build` is enabled and you have multiple Copr build jobs with the `pull_request` trigger, each will trigger its own OSH differential scan. To ensure each scan compares against the right base build, assign matching `identifier` values to each `pull_request`-triggered job and its corresponding `commit`-triggered job. If identifiers are missing or don't match, Packit falls back to the first `commit`-triggered job configured for that branch — which can mean multiple scans end up comparing against the same (possibly wrong) base build.
```

Also, would it be possible not to have a 550-characters-long line? 😅

mfocko (approved):

PR #1172, opened 2026-09-15
Document the `-i` and `--id` alternatives to `--identifier`

betulependule:
It appears that the `-i` and `--id` alternatives to the `--identifier` test argument are missing from the docs.

I believe these alternatives are supported based on the following line of code in `packit-service`:

https://github.com/packit/packit-service/blob/c1f019130b4e81a0c366a4886230a0115967eee8/packit_service/worker/helpers/testing_farm.py#L119

Existing tests also support this belief:

https://github.com/packit/packit-service/blob/c1f019130b4e81a0c366a4886230a0115967eee8/tests/unit/test_testing_farm.py#L2156-L2177


mfocko (approved):
LGTM!

